PR #11, opened 2017-03-21
Specifying UTF-8 encoding

### LINKED ISSUE #9: use 'utf-8' instead of 'ascii'
In 'sigmf-spec.md':

> string: A string of ASCII characters, as defined by the JSON standard.

should be changed to

> string: A string of UTF-8 characters, as defined by the JSON standard.

I'm not familiar with JSON, but from looking at the provided links, it seems like ECMA-404 refers to strings as UTF8 characters. 

> A string is a sequence of zero or more Unicode characters, wrapped in double quotes, using backslash escapes. A character is represented as a single character string. A string is very much like a C or Java string.

Also, UTF8 is probably required as soon as non-english users get involved.

bhilburn:
Addresses Issue #9.

Other changes were an update to the ToC, and an auto-adjust of whitespace to the modified paragraph.

@jdemel @kpreid - Can you confirm this resolves #9?

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -101,7 +101,8 @@ A SigMF recording consists of two files: a SigMF `metadata` file and a
 file contains information that describes the dataset.
 
 1. The metadata and dataset MUST be in separate files.
-2. A metadata file MUST only describe one dataset file.
+2. The metadata file MUST only describe one dataset file.
+3. The metadata file MUST be stored in UTF-8 encoding.
 3. The metadata file MUST have a `.meta` filename extension.
 4. The dataset file MUST have a `.data` filename extension.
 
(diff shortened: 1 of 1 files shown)


kpreid:
If I were being very exact, I would say that the UTF-8 part applies only when the JSON text is stored in a filesystem or other storage that stores bytes without encoding. For an example of not that, a web app which processes SigMF metadata _in JavaScript_ would find this constraint either nonsensical or impractical, because JavaScript works (mostly) with Unicode strings and not bytes-which-are-being-interpreted-explicitly-according-to-some-encoding.

In context of this specification, that means I would put the remark about UTF-8 in the “Files” section instead of the “Metadata Format” section.

Also, there are still two references to “ASCII” that should be simply deleted.

bhilburn:
Oop, good catch on the mention of ASCII. I didn't even think to search for that. And fair point regarding where the UTF-8 spec is made.

PR updated.

bhilburn:
Thanks, @kpreid!

PR #21, opened 2017-03-22
Distinguish between `capture` and its elements.

kpreid:
In order to make it clear that there is an array, and to avoid the misnomer of calling a (JSON) array a (JSON) object, introduce separate sections for `capture` and for the objects inside it. Skip saying "are composed of name/value pairs" for objects, because that's what being an object means.

I think there's room for more precision overall around the use of the words "name/value pair", "object", and so on, but I'll save that for a later more general copyedit or discussion. I'm doing this separately because I repeatedly find it confusing that we currently have "capture object" and "global object" but the value of `core:capture` is _not an object_, and "capture object" has been inconsistently been used in discussion to refer to either the array or its elements.

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -163,7 +163,7 @@ in use, or omitted entirely (including the colon).
    JSON object (i.e., the first character of the file MUST be '{' and the last
    character of the file MUST be '}'. This object is hereafter called the
    `top-level object`.
-3. The top-level object MUST contain three objects named `global`, `capture`,
+3. The top-level object MUST contain three values named `global`, `capture`,
    and `annotations`.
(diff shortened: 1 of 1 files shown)


bhilburn:
Marking this as a bug fix, because @kpreid points out a couple of inconsistencies in our nomenclature.

I just reviewed it and approve. I would like to get one more other than me to look it over, too. Tapping @smunaut this time around to get another set of eyes.

kpreid:
I just noticed that annotations have exactly the same problem in language…

bhilburn:
Thanks, @kpreid. I applied your changes to the Annotation section, as well.

Issue #10, opened 2017-03-21
Annotating Time Series


bastibl:
I think it's quite common to record some kind of time series data together with the actual signal.
For example, I usually record GPS traces. Someone else might record other kinds of sensor data or just log interesting events in a notebook (if someone walked through the experiment, for example).

Intuitively, I would think that this are annotations (also because annotations support latitude and longitude), but the current specification requires to annotate based on samples. The problem is that the data is usually recorded independent from the sample stream and doesn't directly map to a sample index. To me, it feels wrong to do a lot of preprocessing (map time to sample index) to get my raw data into the SigMF format. Apart from that, there might not even be samples recorded at a given point in time (due to gaps between segments). Also, if the segments get large, clock drift might be become a problem.

AFAIS, the only option would be to add the time series as a custom object to global. Like

```
{
    "global": {
        "bastibl:gps": [
            {time1, reading1},
            {time2, reading2}
         ]
    }
}
```

I think it would be nicer to also allow annotations based on time stamps. Currently, annotations seem to be really tuned towards signal identification, i.e. labeling some samples.

smunaut:
I wouldn't allow that in the core spec.  To do anything with those _someone_ will have to map the sample index to timestamp and I'd rather it'd be the recording application rather than force the burden on all the reader, unless you allow all the readers to just ignore those by putting it in an extension.

bastibl:
I don't get why everything has to be mapped to a sample. If you have temperature recording, a picture that is taken at some moment, or a GPS traces, then, for the above mentioned reasons (independent recording, no data available, clock drift), it doesn't make a lot of sense to map it to a sample. Still, I would regard it as an "annotation".

smunaut:
I said "sample index" not sample. It's just a different time unit that's much simpler (i.e. single monotonic integer). There might be no actual sample data for that index.

Logically those infos will need to be displayed or interpreted in relation to the actual data which means they'll need to be time synced. If you're not going to map those annotation to the actual signal recorded, you might as well not put them in the file.

bastibl:
With segments recorded at different sample rates, I don't see how this is a much simpler notation for time. 
Looks like I don't understand the definition of sample index. I thought it is the index of the sample in the file (plus optional offset). But it seems to be an abstract concept for time. Also I wouldn't know how to properly assign a "sample index" to data that is recorded in a gap between segments with different sample rates.
That's too complicated. I'll just use the private namespace.

bhilburn:
[...]

PR #27, opened 2017-03-22
Clarify usage of core:offset

mbr0wn:


bhilburn:
Thanks, @mbr0wn.

Pinging @smunaut for a review, as well. This is obviously a really simple change, but especially with the discussion in #10 I want to make sure we all agree on how we talk about sample indices in the spec :)

smunaut:
The usage I saw for this fields where, as indicated in the commit, when you have a recording split over several files or when you have time synced captures at the same sample rate to indicate the time offset between them.

But that can also be used when splitting / slicing capture data to avoid having to rewrite all the offsets in the file.

All in all it's a general purpose field you can use however you'd like. Readers should just apply the offset.

The only thing I would change is :
* The field should be int64 not uint64
* All sample index should actually be int64 and always fit inside int64 when the offset is applied (so that readers don't have to handle INT64_MAX + INT64_MAX cases)

bhilburn:
Comment from @kpreid:

> That seems reasonable. I feel like there's some clarification that ought to be made (along the lines of "applications should not assume that core:offset means anything in particular other than how sample indexes should be displayed/accepted for reference by users" but more objectively defined) but I can't think of it at the moment.

bhilburn:
Thanks so much for the quick fix, @mbr0wn! =)

PR #30, opened 2017-04-11
Adding `generator` field per #6.

### LINKED ISSUE #6: Annotations: Source is computer or human?
Paraphrased feedback I got here at DARPA's Battle of the ModRecs at DySPAN:

> It would be good to include a field in the annotation segments that define whether or not the segment was created by a program or a human. Right now it looks like people would just use the `comment` name/value for this, but you risk a "sub-standard" being created within that `comment` field since it's so wide-open.

We could do this by adding something to the `core` namespace like `author` or `source-type`, for example.

Thoughts on this idea / concern? In a situation where many different people / groups are producing annotations for the same data set, which they then share, this could be useful.

bhilburn:
Resolves #6 by providing an optional field by which the user can indicate the source of an annotation.

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -254,6 +254,7 @@ the `annotation` object:
 |----|--------------|-------|-----------|
 |`sample_start`|true|uint|The sample index at which this segment takes effect.|
 |`sample_count`|true|uint|The number of samples that this segment applies to. |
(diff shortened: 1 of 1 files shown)


bhilburn:
With no comments disagreeing, I am proceeding with merge.

Issue #8, opened 2017-03-16
Combine date and time values for simplicity and multi-day recordings


kpreid:
I was asked to review the SigMF specification by @bhilburn. It currently says:

> Global Object
>
> … `date` … An ISO-8601 formatted string of the capture date of the recording.
>
> Capture Object
>
> … `time` … An ISO-8601 formatted string indicating the timestamp of the sample index specified by `sample_start`.

I believe that this is likely to be interpreted inconsistently, and will lead to SigMF applications needing to do an unnecessary amount of “calendar” processing (which is easy to get wrong). It is not even clear to me how a recording spanning multiple days would be represented correctly.

I suggest the following change:

* Remove the `date` field from the global object entirely.
* Specify that the `time` field in segments is ISO 8601 combined date and time (as in `2017-03-15T21:11:35Z`).
* While not strictly relevant, I would also recommend requiring the use of the UTC/"Zulu" time zone.

An application wishing to have a date for the recording can simply use the date of the `time` value for the first segment.

bastibl:
I agree that the `date` field should be removed from global and that `time` should be a combined date and time stamp.
In the example metadata file it already is
https://github.com/gnuradio/SigMF/blob/3e373892d0f022826a6b59596b2a5b7b6430135e/example_metadata.py#L19
Maybe it's just a bug in the specification.

Regarding the time zone, we could allow to set one (as in the example file) and assume UTC if none is given.

bhilburn:
Excellent point, @kpreid, and thanks for your thoughts, @bastibl.

The `time` field in the `capture` object is indeed already `datetime`, as @bastibl pointed out, and its name is something of a misnomer.

I am actually a fan of the idea of specifying UTC Zulu. Having to deal with Daylight Savings, for example, is a complete nightmare. It's not even consistent in the USA - it's observed by all states except Arizona and Hawaii, unless you are on tribal lands in those states, in which case it is observed, and it is not observed in territories. Specifying UTC Zulu, which doesn't change, seems much simpler.

Proposal:

* Remove `date` from `global` object
* Change name of `time` in `capture` object to `datetime`
* Require UTC Zulu time

Thoughts / feedback?

kpreid:
I was less concerned about the _name_ of the field and more about precision in what it is defined to contain. As long as the spec is not ambiguous, people will get the idea quickly from examples. I have little opinion on using `time` vs. `datetime`.

[...]

PR #31, opened 2017-04-11
Addressing #8 - making `datetime` clearer in `capture` segments.

bhilburn:
Per #8, made the following changes:

* Remove date from global object
* Change name of time in capture object to datetime
* Require UTC Zulu time

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -241,13 +241,29 @@ capture segment objects:
 
 ###### The `datetime` Pair
 
-This name/value pair must be an ISO-8601 string of the form
-`YYYY-MM-DDTHH:MM:SS.SSSZ`, where the `T` is the delimiter between the date and
-the time, and the `Z` indicates the UTC/Zulu timezone. All timestamps must be in
-the UTC/Zulu timezone, and use the 24-hour clock.
+This name/value pair must be an ISO-8601 string, as defined by [RFC
+3339](https://www.ietf.org/rfc/rfc3339.txt), where the only allowed
+`time-offset` is `Z`, indicating the UTC/Zulu timezone. The ABNF description is:
 
-Any number of digits of precision may be used in the fractional seconds portion
-of the timestamp.
+   date-fullyear   = 4DIGIT
+   date-month      = 2DIGIT  ; 01-12
+   date-mday       = 2DIGIT  ; 01-28, 01-29, 01-30, 01-31 based on
+                             ; month/year
+   time-hour       = 2DIGIT  ; 00-23
+   time-minute     = 2DIGIT  ; 00-59
+   time-second     = 2DIGIT  ; 00-58, 00-59, 00-60 based on leap second
+                             ; rules
+   time-secfrac    = "." 1*DIGIT
+   time-offset     = "Z"
+
+   partial-time    = time-hour ":" time-minute ":" time-second [time-secfrac]
+   full-date       = date-fullyear "-" date-month "-" date-mday
+   full-time       = partial-time time-offset
+
+   date-time       = full-date "T" full-time
+
+Thus, timestamps take the form of `YYYY-MM-DDTHH:MM:SS.SSSZ`, where any number
+of digits for fractional seconds is permitted.
 
 #### Annotation Array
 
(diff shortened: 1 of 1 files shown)


sigmf-spec.md
```diff
@@ -231,7 +230,7 @@ the `capture` object:
 |`sample_start`|true|uint|The sample index at which this segment takes effect.|
 |`frequency`|false|double|The center frequency of the signal in Hz.|
 |`sample_rate`|false|double|The sample rate of the signal in Samples per Second.|
-|`time`|false|string|An ISO-8601 formatted string indicating the timestamp of the sample index specified by `sample_start`.|
+|`datetime`|false|string|An ISO-8601 formatted string indicating the timestamp of the sample index specified by `sample_start`, using the UTC/Zulu timezone.|
```
kpreid: I'd like this even better if it said exactly which of the string formats ISO 8601 defines is allowed here, rather than the non-specific "ISO-8601 formatted string".

I don't know what the proper wording is and ISO 8601 is not freely available so I can't be more specific. Wikipedia used the phrase “combined date and time”. One could write a template like “`YYYY-MM-...`” and rely on ISO 8601 to disambiguate the details.

bhilburn:
Okay, so as @kpreid pointed out, it would be good to specify an actual format as ISO-8601 does indeed allow for a lot. I'm partial to this:

```
YYYY-MM-DDTHH:MM:SS.SSSZ
```
Where the `T` and the `Z` are always present to indicate the delimeter between date and time and the ZULU timezone, respectively. ISO-8601 and RFC 3339 (the subset of ISO-8601 used for the Internet) both allow for fractional seconds (as you can see, above), but do not limit the number of digits that can be used, there. 

So, the obvious question, then, is should we specify a maximum number of digits? If we don't, we may complicate application implementations.

Thoughts?

kpreid:
Thoughts:

Milliseconds, microseconds, and nanoseconds are _all_ popular choices for minimum time resolution, unfortunately.

Defining the precision matters only for _editor_ applications. Readers can (and will) simply discard digits they have no use for; writers can simply write as many digits as they have available.

How much do we care about guaranteeing that an editor preserves the metadata exactly? I have the feeling that there are much bigger problems than this one lurking, but I don't have one to point at offhand.

bhilburn:
Okay, I updated this PR with @kpreid's feedback. 

Thoughts / comments?

sigmf-spec.md
```diff
@@ -238,7 +237,17 @@ capture segment objects:
...
 |`sample_start`|true|uint|The sample index at which this segment takes effect.|
 |`frequency`|false|double|The center frequency of the signal in Hz.|
-|`time`|false|string|An ISO-8601 formatted string indicating the timestamp of the sample index specified by `sample_start`.|
+|`datetime`|false|string|An ISO-8601 string indicating the timestamp of the sample index specified by `sample_start`. More details, below.|
+
+###### The `datetime` Pair
+
+This name/value pair must be an ISO-8601 string of the form
+`YYYY-MM-DDTHH:MM:SS.SSSZ`, where the `T` is the delimiter between the date and
+the time, and the `Z` indicates the UTC/Zulu timezone. All timestamps must be in
```
kpreid: Defining the meanings of T and Z seems beside the point whereas the clarification that is most needed (for anyone who doesn't already find it obvious) is that `T` and `Z` are literal text. Maybe write it out with markup for metasyntactic elements, like <code><var>YYYY</var>-<var>MM</var>-<var>DD</var>T<var>HH</var>:<var>MM</var>:<var>SS</var>.<var>SSS</var>Z</var></code>.

    This name/value pair must be an ISO-8601 string of the form
    <code><var>YYYY</var>-<var>MM</var>-<var>DD</var>T<var>HH</var>:<var>MM</var>:<var>SS</var>.<var>SSS</var>Z</var></code>.
    All timestamps must be in the UTC/Zulu timezone, and use the 24-hour clock.

bhilburn:
Took another shot at this using ABNF. Have a look and comment!

bhilburn:
@kpreid - If you can think of any ways to simplify this, feel free to create an issue (or even better, submit a PR)! Otherwise, we'll roll with this for now. Thanks for your input!

Issue #17, opened 2017-03-21
Omitting namespace strings?


bhilburn:
The spec currently says:

> The format of the name/value pairs is: 
>
> `"<namespace:>name": value,`
>
> Where the namespace string should be replaced with the name of the namespace in use, or omitted entirely (including the colon).

I think that was originally written before we had completely fleshed out the concept of namespaces. Do we really want to allow omission of the namespace string? Allowing the omission of `"core"`, for example, actually means that it's not semantic, which isn't something I agree with.

smunaut:
should be required always IMHO.

bhilburn:
Okay, see #32 and review, please!

bhilburn:
Closed by #32

PR #32, opened 2017-04-11
`namespace` is now required, per #17

bhilburn:
The `namespace` string is no longer optional.


djanderson:
This commit should also update the wording of the spec if it's in fact not optional now.

bhilburn:
Not seeing any dissent, so merging.

Issue #18, opened 2017-03-21
Have only one way to write little-endian


kpreid:
I was asked to review the SigMF specification by @bhilburn. It currently says:

> All of these must be explicitly specified, except for endianness which is assumed to be little-endian unless specified otherwise.
>
> ...
> 
> `<r|c>``<f|i|u>``<8|16|32>``[<_le|_be>]`

This seems like a pointless degree of freedom, which creates the opportunity for confusion such as taking no-suffix as meaning platform native or big endianness, particularly if an implementor works from examples rather than spec.

I suggest making `<_le|_be>` non-optional. (Also fix the markup that is showing <code>`</code> characters here while you're at it.)

bhilburn:
Hm, yeah, that markdown is broken.

Regardless, it's a fair point. I believe we originally did this in an effort to keep things as simple as possible, but I agree it does open the door to confusion where we could easily avoid it.

I'm on-board with the change. Opinions from anyone else?

bhilburn:
Okay, PR is up at #33!

bhilburn:
Closed per #33's merge.

Issue #5, opened 2017-03-01
Specify Floating-Point Standard


bhilburn:
Comment from reviewer:

> Note that the current SigMF does not specify -which- floating point format to use. For 32 bit, IEEE-754 would seem sensible, and presumably the half-precision format from IEEE754-2008 ?

It's a good point. Agree / disagree with the suggestion?

smunaut:
I didn't remember we spec'd support for half-float at all ..

IEEE-754 makes sense, but I'm wondering if we shouldn't allow for undefined behavior for non-finite values and non-normalized ones. After all we expect physical values here and using -ffast-math makes sense a lot of the time. Just a thought ...

bhilburn:
@smunaut raises an interesting point. Technically, as our datatypes are defined, someone could actually use `cf8` and it would be valid. This is item `#1`.

Item `#2` is @smunaut's point about finite/normalized values. Thoughts, here? Guaranteeing compatibility with things like `--ffast-math` is nice, but is that moving too far in the direction of recommended toolflows? Do we preclude other tools by *not* allowing for non-finite/non-normalized values?

bhilburn:
Okay, so, fixing this is going to require a change to the general format of the `datatype` section, as we need to limit the floating-point options.

Proposal: we only allow single-precision IEEE 754 floating point (32-bit floats) samples, and do not allow for half-precision (16-bit floats).

Question: Should we add double-precision to the standard? I'm leaning towards `no`, as that may become a burden to becoming `SigMF Compliant`. This could be done through an extension, instead. Thoughts?

Question: I would like to get some more input on @smunaut's point on finite/normalized values.

bhilburn:
Unless anyone has further input on my question, above, I'm going to move forward with the proposal as I stated it.

bhilburn:
Okay, #43 is up for this.

bhilburn:
Closing this as the discussion has moved to #43.

PR #33, opened 2017-04-11
Specifying endianess now required per #18.

bhilburn:
No longer making an assumption about `_le` if none is specified.


bhilburn:
This should get merged before #5 is addressed to prevent conflicts.

bhilburn:
Merging with @kpreid's approval.

Issue #22, opened 2017-03-22
Clarify effect of omitted segment metadata


kpreid:
Given that

* segments have a `core:sample_start` property but not any notion of end-point, and
* individual segment metadata names are optional,

it seems likely that applications might interpret it as that if some metadata is not specified it is carried over from the previous segment (at least in the case of `frequency` and `sample_rate`) — the model of "when a new segment is seen, assign the values in it to the current-state variables".

I request that this be clarified.

Further, I suggest it be specified that such metadata is _not_ carried over. This will allow straightforward representation of regions of a recording where we do not know some information; for example, the center frequency was changed and it takes some time for the LO to resettle, so we make a segment spanning that event which has no `frequency`.

bhilburn:
@kpreid has a good point, here, and allowing fields to be carried over could result in:

(a) having to walk backwards to gather all of the attributes that apply to the current segment
(b) ambiguity in whether an attribute applies to the current segment unless you specify the negative, which is even more confusing.

I have submitted a PR for this. Please review #34!

bhilburn:
Resolved by #34.

PR #34, opened 2017-04-11
Prohibiting the carrying over of metadata between segments.

bhilburn:
I tried to do this in the simplest way possible by stating this in a single location. Open to feedback on better places to put this or state it more clearly.

This addresses #22.


bhilburn:
Merging as I'm not seeing further discussion and we have @kpreid's review approval.

PR #37, opened 2017-04-22
spec: Annotation frequencies are in (complex) base band

schneider42:
The spec is not explicit about if the frequencies of an annotation are in passband or baseband.

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -259,8 +259,8 @@ the `annotation` object:
 |`sample_start`|true|uint|The sample index at which this segment takes effect.|
 |`sample_count`|true|uint|The number of samples that this segment applies to. |
 |`comment`|false|string|A human-readable comment.|
-|`freq_lower_edge`|false|double|The lower edge of the frequency band of a signal feature that this annotation describes.|
-|`freq_upper_edge`|false|double|The upper edge of the frequency band of a signal feature that this annotation describes. |
+|`freq_lower_edge`|false|double|The lower edge (in the baseband) of the frequency band of a signal feature that this annotation describes.|
+|`freq_upper_edge`|false|double|The upper edge (in the baseband) of the frequency band of a signal feature that this annotation describes.|
 |`latitude`|false|need a standard?| |
 |`longitude`|false|need a standard?| |
 


kpreid:
I'm not sure if "in the baseband" is the clearest way to say it but this is a good clarification to make!

schneider42:
I wasn't sure either. I just didn't want make the sentence / entry in the table much longer.

Any suggestions?

kpreid:
Best short phrase that I have thought of is “(relative to the center frequency of the dataset)” but I don't find that to be great because we actually mean it is _absolute_ with respect to the _actual recorded samples_ and center frequency is added-on metadata.

I was concerned about using the term “baseband” because it makes me think about frequencies of _a particular modulated signal_ — "carrier frequency = 0 Hz" — whereas recordings are likely to have a center/zero-Hz point that is _not the carrier._ But I'm not very familiar with the precise use of “baseband” so perhaps there's nothing to worry about.

bhilburn:
@schneider42 - Excellent point! Yes, this is definitely something that needs to be clarified.

@kpreid has a fair point about the use of `baseband`, but I think we are probably okay using it in the way you have written, @schneider42. 

I think I actually would have gone the opposite direction, here, and used the RF frequency rather than the baseband frequency as I think it makes looking at a single annotation segment and immediately grokking it a bit easier.

I would be interested to get @djanderson's opinion, here.

kpreid:
Hmmmm. I hadn't actually thought about the question of which is _better_. And when I think about it, I notice that **the center frequency can change** — in fact, it can change within the span of a single annotation. That means that there are several different choices of interpretation:

1. Annotations are written in RF frequency; annotations are rectangles in RF/time space.
2. Annotations are written in baseband frequency, relative to the center frequency at the beginning point of the annotation, and one of:
    1. the annotation is a rectangle in (absolute/RF) × (time) space, so it is non-rectangular in (baseband) × (time) space.
    2. the annotation is a rectangle in (baseband frequency) × (time) space and non-rectangular in (RF) × (time) space.
4. It is an error for annotation segments to span capture segments that have different center frequencies.

The “rectangular in baseband” options would be easiest for a graphical viewer application to display, but that's not a very good criterion; we should be making the choice based on which is _least likely to result in recordings with bad metadata._ I think perhaps option 1 is best, because it offers no “relative to what?” questions/mistakes.

Except that if we choose option 1 we must decide what to do with recordings which have no center frequency (either because it is unknown or because it is synthesized rather than recorded off the air and thus never had one). We could declare that in that case they are relative — the center frequency is zero for this purpose only.

bhilburn:
Moving over to #52.

PR #38, opened 2017-04-28
Add setuptools setup.py and setup.cfg

djanderson:
These allow for installing the SigMF utilities via `pip install -e .` in
the source directory, or directly from the git repo.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[wheel]
+universal = 1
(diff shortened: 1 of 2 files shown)


a5552033:
I am sorry for ask that question about gr-ltetrigger. .I want to know how to get the LTE frame like ../test_frames/lte_frame_50prb_cellid_125 by gr-ltetrigger?
Hope your reply!
Peggy

bhilburn:
@djanderson - Great call! Thanks so much for submitting a PR to get us going, here =)

PR #39, opened 2017-05-01
Add test for and fix SigMFFile default constructor

djanderson:
The SigMFFile class contained the following, which makes it appear you wanted to allow initializing a SigMFFile with no input and have it use some sane default schema:

```
    def __init__(
            self,
            metadata=None,
            data_file=None,
            global_info=None,
    ):
        self.version = None
        self.schema = None
        if metadata is None:
            self._metadata = get_default_metadata(self.get_schema())
        [...]
```

But `__init__` as written has a circular dependency on `self._metadata`, so I get the following:

```
>>> from sigmf import SigMFFile
>>> f = SigMFFile()
  File "/home/dja/dev/SigMF/sigmf/sigmffile.py", line 80, in __init__
    self._metadata = get_default_metadata(self.get_schema())
  File "/home/dja/dev/SigMF/sigmf/sigmffile.py", line 113, in get_schema
    current_metadata_version = self.get_global_info().get(self.VERSION_KEY)
  File "/home/dja/dev/SigMF/sigmf/sigmffile.py", line 132, in get_global_info
    return self._metadata.get(self.GLOBAL_KEY, {})
AttributeError: 'SigMFFile' object has no attribute '_metadata'
```

This commit adds a unit test for the default constructor and then fixes the circular dependency.

### DIFF
--- a/sigmf/sigmffile.py
+++ b/sigmf/sigmffile.py
@@ -129,7 +129,10 @@ class SigMFFile(object):
         """
         Returns a dictionary with all the global info.
         """
-        return self._metadata.get(self.GLOBAL_KEY, {})
+        try:
+            return self._metadata.get(self.GLOBAL_KEY, {})
+        except AttributeError:
+            return {}
 
     def set_global_field(self, key, value):
         """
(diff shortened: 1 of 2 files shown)


bhilburn:
@djanderson - Nicely done, and thanks for the contribution!

PR #40, opened 2017-05-02
Update example metadata for PR #32

djanderson:
The spec states "Each capture segment must contain a core:sample_start
name/value pair", and PR #32 makes the namespace required, so the
example should reflect that.


bhilburn:
Thanks for the contribution, @djanderson! It is much appreciated.

PR #43, opened 2017-05-03
Specifying that only IEEE-754 single-precision is permitted.

bhilburn:
Specifies that only single-precision floats are permitted. I'm sure there is a more elegant way to state this within the spec, but it wasn't clear to me immediately how to go about it. Let me know if anyone can think of a better way to state this?

This addresses #5.

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -140,23 +140,28 @@ four orthogonal characteristics of sample data: complex or real, floating-point
 or integer, bit-width, and endianness. The following ABNF rules specify the
 dataset formats defined in the SigMF `core` namespace:
 
-    dataset-format = (real / complex) type endianness
+```abnf
+    dataset-format = (real / complex) ((type endianness) / byte)
 
     real = "r"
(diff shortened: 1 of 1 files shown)


kpreid:
What I would do is specify the combinations explicitly: say that only `f32 i8 i16 i32 u8 u16 u32` are supported. This is not too long, does not miss mentioning anything, and implementations will most likely have separate code for all of those choices.

Hm. Not the same issue, but as currently written, `cu8` is not legal and it has to be `cu8_be` or `_cu8_le`. That seems unfortunate.

bhilburn:
Okay, so I actually went with ABNF to define this. Have a look and comment.

@kpreid - I didn't solve your problem of `cu8` needing to be `cu8_be` or `cu8_le` in the update, but you have a good point, there. Will have a look at how to make that possible.

bhilburn:
Okay, fixed the markup from my previous commit and fixed the issue with LE / BE bytes that @kpreid pointed out.

Issue #15, opened 2017-03-21
Define a standard single-file format


kpreid:
I was asked to review the SigMF specification by @bhilburn.

Data formats consisting of multiple files are frequently awkward to work with; for example, when downloading them from a web site. People will likely decide to distribute them in archives instead.

Therefore, I propose that the specification preemptively define a simple single-file format, which straightforwardly contains the metadata and data files. There are a lot of possibilities here, so here are some suggested constraints:

* The format should be a ZIP archive. (Commonly used as a multi-file container; precedent in e.g. JAR and EPUB files. Can be written in streaming fashion (no seeks).)
* The single-file format should have a unique filename extension such as `.sigmf`, even if it is itself a standard archive format such as ZIP. (Same argument as in #14, and should be coordinated with that.)
* If the format is a standard archive format, it should be required to have specific filenames (nontrivial relative, or absolute, paths prohibited).
* No other contained files besides the SigMF metadata and data files are permitted. (This ensures that a round-trip between the two-file and single-file formats does not lose anything; thus avoiding introducing any more complexity than necessary to support single-file.)
* Implementations SHOULD avoid attempting to compress the data file, since doing so is highly unlikely to give any advantage.

bhilburn:
I really like this suggestion a lot, and you're right - if we don't specifically call this out, there are going to be SigMF recordings floating around as `zip`, `rar`, `tar`, `tar.gz`, `tar.gz2`, `7-zipped directories`, and a dozen other things.

I agree with every bullet in @kpreid's top-level comment, but do want to follow-up on one item:

> If the format is a standard archive format, it should be required to have specific filenames (nontrivial relative, or absolute, paths prohibited).

Is the suggestion, here, that the filenames of the metadata and dataset files within the recording be fixed?

kpreid:
The most important part is: pathnames referring to some other directory, relatively or absolutely, are prohibited; only filenames are allowed.

Whether the filenames themselves are fixed has a tradeoff:

1. If we require a specific name (`the.data, the.meta`, modulo #14), then things are simple and strict but using a regular unarchiver generates unhelpful-for-distinguishing names (though often in an automatically-created directory named after the archive, which seems like a decent result).
2. If we allow the name prefix to vary (`foo.sigmf` may contain `foo.data, foo.meta` and so on), then this is convenient for unarchiving, but means that renaming a `.sigmf` container potentially leaves an old, “wrong”, name behind to surprise you later.

[...]

PR #44, opened 2017-05-03
Creating the `.sigmf` archive format

bhilburn:
This creates a SigMF archive format, per the discussion in #15.

I fleshed out some ideas not previously discussed in #15, including combining multiple recordings into a single SigMF archive, so please review.

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -109,17 +109,16 @@ file contains information that describes the dataset.
 #### Archive Format
 
 The metadata and dataset files that comprise a SigMF recording may be combined
-into a file archive. Multiple SigMF recordings may be stored in a single
-archive.
+into a file archive.
 
 1. The archive MUST use the `tar` archive format, as specified by POSIX.1-2001.
 2. The archive file's filename extension MUST be `.sigmf`.
-3. Metadata and dataset files MUST NOT be in the top-level directory of the
-   archive.
-4. Each SigMF recording (comprised of a metadata file and a dataset file) MUST
-   be in separate directories.
-5. The metadata and dataset files within a directory must have the same
-   filenames as the directory itself.
+3. The archive MUST contain the following files: for each contained recording
+   with some name given here meta-syntactically as `N`, files named `N` (a
+   directory), `N/N.sigmf-meta`, and `N/N.sigmf-data`.
+4. The archive MUST NOT contain any other files unless their pathnames begin
+   with `N/N`. for some `N` which has `.sigmf-meta` and `.sigmf-data` files as
+   described above.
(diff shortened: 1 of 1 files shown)


sigmf-spec.md
```diff
@@ -106,6 +106,21 @@ file contains information that describes the dataset.
 3. The metadata file MUST have a `.meta` filename extension.
 4. The dataset file MUST have a `.data` filename extension.
 
+#### Archive Format
+
+The metadata and dataset files that comprise a SigMF recording may be combined
+into a file archive. Multiple SigMF recordings may be stored in a single
+archive.
```
kpreid: I think we should discuss the idea of combining multiple recordings separately before putting it in the spec. It is a UI/API complication for anything that wants to read a SigMF recording because you then need more than a file-picker, and people will be tempted to create little languages for representing pathnames-plus-item-in-the-archive, which gets messy.

sigmf-spec.md
```diff
@@ -106,6 +106,21 @@ file contains information that describes the dataset.
...
+
+1. The archive MUST use the `tar` archive format, as specified by POSIX.1-2001.
+2. The archive file's filename extension MUST be `.sigmf`.
+3. Metadata and dataset files MUST NOT be in the top-level directory of the
+   archive.
+4. Each SigMF recording (comprised of a metadata file and a dataset file) MUST
+   be in separate directories.
+5. The metadata and dataset files within a directory must have the same
+   filenames as the directory itself.
```
kpreid: I don't like how point 3 is written negatively; it does not rule out that you could have two levels of directories (which is another complication), or using absolute paths (which would be bad).

I would suggest something to the effect of (I don't know what language POSIX uses to describe tar files):

3. The archive MUST contain the following files: for each contained recording with some name given here metasyntactically as `N`, files named `N` (a directory), `N/N.sigmf-meta`, and `N/N.sigmf-data`.

4. The archive MUST NOT contain any other files unless their pathnames begin with `N/N.` for some `N` which has `.sigmf-meta` and `.sigmf-data` files as described above.

(The item 4 is to allow for extensions without allowing for arbitrary junk. More minimally, we could say that for now no other files are allowed.)

sigmf-spec.md
bhilburn: Good point - this is perhaps a little more complex than I first anticipated.

sigmf-spec.md
bhilburn: I really like your suggestion, @kpreid. Nice work =)

bhilburn:
Okay, PR updated with the comments from @kpreid, above.

Issue #14, opened 2017-03-21
Use more specific filename extensions


kpreid:
I was asked to review the SigMF specification by @bhilburn. It currently says:

> 3. The metadata file MUST have a `.meta` filename extension.
> 4. The dataset file MUST have a `.data` filename extension.

I recommend, instead, specifying a (closer to) globally unique pair of filename extensions. This allows uses such as:

* Registering editor/viewer applications in traditional desktop environments.
* An application deciding how to interpret a file in one of several possible formats by looking at the extension.
* A user web searching to find out what format the files they have are in.
* Registering in web server MIME type databases.

I suggest either `.sigmfm` and `.sigmfd` or `.sigmfmeta` and `.sigmfdata`, the choice depending on a weighting of clarity vs. shortness.

(Starting them with a common prefix also assures that the two files will sort next to each other if they are accompanied by more files with other extensions.)

bhilburn:
How do folks feel about hyphenated extensions?

`.sigmf-meta` and `.sigmf-data`

kpreid:
I'm not aware of any particularly strong reason to choose or avoid them. I can think of several vague FUDy reasons to avoid them: I imagine there might, somewhere in the world, be a thing that thinks filename extensions can only contain letters/numbers. It's one character longer. It might be taken as two tokens in a keyword search.

bhilburn:
Okay, #45 is up for this!

bhilburn:
Fixed by #45.

PR #45, opened 2017-05-03
Making filename extensions more unique.

bhilburn:
Okay, this makes the filename extension changes discussed in #14.

Pretty simple change, but please review!

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -103,9 +103,9 @@ file contains information that describes the dataset.
 1. The metadata and dataset MUST be in separate files.
 2. The metadata file MUST only describe one dataset file.
 3. The metadata file MUST be stored in UTF-8 encoding.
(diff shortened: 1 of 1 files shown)


sigmf-spec.md
```diff
@@ -103,8 +103,8 @@ file contains information that describes the dataset.
 1. The metadata and dataset MUST be in separate files.
 2. The metadata file MUST only describe one dataset file.
 3. The metadata file MUST be stored in UTF-8 encoding.
-3. The metadata file MUST have a `.meta` filename extension.
-4. The dataset file MUST have a `.data` filename extension.
+3. The metadata file MUST have a `.sigmf-meta` filename extension.
```
kpreid: fix list item numbering (doesn't actually matter on display, but might confuse people)

bhilburn: Oops! Thanks for the catch.

Issue #28, opened 2017-03-22
Changing Sample Rates


bhilburn:
Currently, the spec allows for the sample rate to change within a recording by writing a new capture segment and specifying the new rate.

In the discussion in #10, both @smunaut and @bastibl raised this as an issue:

> TBH the case of multiple sample rate I just don't think should be allowed. sampe_rate is a fundamental property of how to interpret the data and shouldn't be changed in the middle of a stream (like data type for instance).

I believe the original goal of the spec was to be as flexible as possible, but allowing this does significantly complicate other things, including the implementation of reader applications.

I'd like to get others' opinions on this matter.

kpreid:
Changing sample rate is awkward because if you're not doing something which doesn't need to know it (e.g. a waterfall display is calculated the same and just has a different scale) then you get to reconfigure your DSP, and it doesn't make sense to have a window of samples spanning the change point. So there's a sense in which it _might as well be different data files._

On the other hand, if one has an interactive receiver application which is continuously recording and allows the sample rate to be changed, a mid-file rate change is an accurate recording of what happened. (But such an application would probably want to create Reasonable-Sized Chunks rather than one arbitrarily large file, anyway, so just an early split.)

I would be tempted to declare changing sample rate a feature which compliant applications are not required to support.

smunaut:
not so "accurate"  ... even URSP doing stream tags don't properly tag at the exact sample where the rate changed.

And yeah, I'd agree that such a recording application is pretty much the only case I see where this would happen, but as you said, splitting early is probably the better option here.

So I'd definitely leave support for that kind of things to extensions.

bhilburn:
Okay, it sounds like the consensus is that while this shouldn't necessarily be outright prohibited, it shouldn't be a requirement of the core spec.

So, next question: where should the `sample_rate` name/value pair be, then? If we put it in the `capture` object, per #22/#34 it will have to be re-stated in every capture segment even though it isn't changing. We could put it in the global object, where it would make sense in conjunction with `datatype`.

Thoughts?

bhilburn:
Okay, #48 is up, which makes the changes discussed here.

bhilburn:
Addressed by merging #48.

PR #48, opened 2017-05-03
Making `sample_rate` constant in a recording.

bhilburn:
Okay, `sample_rate` is now defined in the `global` object, and is thus constant for a recording.

This is the result of discussions in #28


sigmf-spec.md
```diff
@@ -206,6 +206,7 @@ the `global` object:
 |----|--------------|-------|-----------|
 |`datatype`|true|string|The format of the stored samples in the dataset file. Its value must be a valid SigMF dataset format type string.|
 |`datapath`|true|string|The filepath to the dataset file described by the SigMF file. The path can be absolute or relative.|
+|`sample_rate`|false|double|The sample rate of the signal in Samples per Second.|
```
kpreid: While we're here: `samples per second`, no capitalization.

PR #51, opened 2017-05-12
Allow setting fields that aren't defined in schema

djanderson:
The spec says anything outside the defined core fields should be
acceptable and ignored. The code as fails when adding a previously
undefined global field. This commit adds a unit test for this case and
implements the correct behavior.

Also fix a JSON notation issue in test_validation.py.


bhilburn:
Thanks for another great contribution, @djanderson! =)

PR #52, opened 2017-05-18
Specifying Annotation Frequency Bands

bhilburn:
This includes @schneider42's work from #37, with another commit from me.

Thoughts on this attempt at rewording & specifying, @kpreid and @schneider42?

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -278,10 +278,16 @@ annotation segment objects:
 |`sample_count`|true|uint|The number of samples that this segment applies to. |
 |`generator`|false|string|Human-readable name of the entity that created this annotation.|
 |`comment`|false|string|A human-readable comment.|
-|`freq_lower_edge`|false|double|The true frequency value (at RF, if applicable) of the lower edge of the feature described by this annotation.|
-|`freq_upper_edge`|false|double|The true frequency value (at RF, if applicable) of the upper edge of the feature described by this annotation.|
-|`latitude`|false|need a standard?| |
-|`longitude`|false|need a standard?| |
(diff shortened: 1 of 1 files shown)


sigmf-spec.md
```diff
@@ -278,8 +278,8 @@ annotation segment objects:
 |`sample_count`|true|uint|The number of samples that this segment applies to. |
 |`generator`|false|string|Human-readable name of the entity that created this annotation.|
 |`comment`|false|string|A human-readable comment.|
-|`freq_lower_edge`|false|double|The lower edge of the frequency band of a signal feature that this annotation describes.|
-|`freq_upper_edge`|false|double|The upper edge of the frequency band of a signal feature that this annotation describes. |
+|`freq_lower_edge`|false|double|The true frequency value (at RF, if applicable) of the lower edge of the feature described by this annotation.|
```
kpreid: “true” and “if applicable” are both too vague.

bhilburn: I don't disagree; this was my shot at trying to get a discussion going about the best words to use. This one is a bit tricky to make clear while also being succinct.

 @schneider42 @kpreid - How about going with the simplest form possible?
```
The absolute frequency of the lower edge of the feature described by this annotation.
```

Thoughts? If you don't care for it, do you have suggestions for other forms?

bhilburn:
Okay, I went back and forth on this one a lot.  Defining the annotations relative to the current `center` frequency (or baseband of none exists) is simpler because there isn't the exception case. But, the fact that the center frequency can change across annotation segments makes that really complicated, unless we add another rule prohibiting that.

In the end, I went with basically exactly what @kpreid suggested: the annotations are at RF is possible, otherwise at baseband. I'm not super satisfied with this, but I don't have a better solution.

PR #53, opened 2017-05-18
Specifying format of "text" fields.

### LINKED ISSUE #7: Be explicit about the format of all “text” fields
I was asked to review the SigMF specification by @bhilburn. This is my first detailed writeup of an issue.

There are several fields in the metadata definition which are said to be “text” or “human-readable”. I recommend that it be specified explicitly and exactly what sort of text. Software and data authors may otherwise make incompatible interpretations, consciously or accidentally, such as:

1. The text is parsed as HTML or some subset thereof.
2. The text is plain text but whitespace is not significant. (That is, the result of putting the text through a HTML entity escaper but not processing whitespace.)
3. The text is plain text and newlines are significant but multiple spaces are not (as in Markdown, loosely speaking).
4. The text is plain text and all whitespace is significant (as in `<pre>`).

(The specific examples here are mostly about displaying SigMF metadata in web browsers, but this will presumably be a very common application anyway.)

I would recommend choosing either interpretation 1 or interpretation 4. Interpretation 4 has the advantage of simplicity; Interpretation 1 has the advantage of being able to include hyperlinks to additional documentation, license documents, authors' sites, catalogs of related files, and so on.

The currently-defined fields for which this clarification should be made are global `description`, `author`, `license`, `hw` and annotation `comment`.

bhilburn:
This partially resolves #7.

This describes the format of "text" fields in the standard, unless otherwise specified.

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -79,6 +79,10 @@ interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
 JSON keywords are used as defined in [ECMA-404](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf).
 
+Fields defined as "human-readable", a "string", or simply as "text" shall be
+treated as plaintext ASCII where whitespace is respected, unless otherwise
+specified.
+
 ## Specification
 
 The SigMF specification fundamentally describes two types of information:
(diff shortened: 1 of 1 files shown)


sigmf-spec.md
```diff
@@ -79,6 +79,10 @@ interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
 JSON keywords are used as defined in [ECMA-404](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf).
 
+Fields defined as "human-readable", a "string", or simply as "text" shall be
+treated as plaintext ASCII where whitespace is respected, unless otherwise
```
kpreid: Referring to “ASCII” is not just over-constraining but incorrect here. The contents of a JSON document are not bytes to be interpreted via an encoding — they are characters.

I would suggest “…shall be treated as plain text where all whitespace is significant, …”

bhilburn: @kpreid - You're right. ECMA-404 specifies `Unicode`, not a specific encoding. Good catch.

I'm good with your suggestion.

bhilburn:
Okay, updated per the change discussion, above.

Issue #25, opened 2017-03-22
Make license information machine-readable when possible


kpreid:
If the license information is not, or not exclusively, free-form text (with potentially many variations on references to a single license) but contains some kind of formally specified reference to the license, then it is possible for automated systems to determine whether they can process a dataset without violating its license.

An archive of datasets might want to display publicly only datasets with compatible attribution requirements, or allow searching with license as a criterion.

The results of such a search could be used as input to some process which combines many datasets (machine learning, distributed receivers, propagation analysis…), producing a derived work without any person needing to evaluate the individual inputs.

#20 may be relevant; I'm sure there are already-existent vocabularies for license metadata but I am not familiar with anything in specific.

mbr0wn:
Yeah, this is worth investigating.

bhilburn:
This is a really good point, although I'm not terribly sure how to proceed.

Perhaps, for now, we can just add a recommendation of this to the spec regarding the license field. Unless anyone has good ideas, here, I'll proceed with that plan.

kpreid:
A simple from-scratch strategy might be "must be a URL for the license document; when possible, one defined by the license author rather than specific to this dataset".

bhilburn:
Okay, #54 is up for this!

bhilburn:
Fixed by merge of #54!

PR #54, opened 2017-05-18
Updating license field description per discussion in #25.

bhilburn:
This addresses #25. The goal is to make it possible for applications to auto-parse the license document if it is known (as Github and Gitlab do, for example).

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -230,7 +230,7 @@ the `global` object:
 |`offset`|false|uint64|The index number of the first sample in the dataset. This value defaults to zero. Typically used when a recording is split over multiple files.|
 |`description`|false|string|A text description of the SigMF recording.|
 |`author`|false |string|The author's name (and optionally e-mail address).|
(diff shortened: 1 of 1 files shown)


sigmf-spec.md
```diff
@@ -228,7 +228,7 @@ the `global` object:
 |`description`|false|string|A text description of the SigMF recording.|
 |`author`|false |string|The author's name (and optionally e-mail address).|
 |`date`|false|string|An ISO-8601 formatted string of the capture date of the recording.|
-|`license`|false|string|The license under which the recording is offered.|
+|`license`|false|string|A URL for the license document under which the recording is offered; when possible, use the canonical document provided by the license author.|
```
kpreid: “provided by the license author or, failing that, a well-known one.”

For example, the traditional open source licenses were first used on individual projects and then people reused them without that being the intent of the author.

bhilburn: Ah, fair point. I'm good with that.

bhilburn:
Updated per change discussion, above.

Issue #56, opened 2017-06-01
annotations vs annotation, captures vs capture


djanderson:
The spec as currently written says 

> The `capture` value is an array of _capture segment objects_

and

> The `annotation` value is an array of _annotation segment objects_

However, the examples and schema appear to use the key `annotations` instead of `annotation`.

https://github.com/gnuradio/SigMF/blob/master/example_metadata.py#L26
https://github.com/gnuradio/SigMF/blob/master/sigmf/schema.json#L84

First thought: capture and annotation should either both be singular or both be plural. Let me know which it is and I'll do a pull request.

I actually like plural better since they are arrays of those objects. For example, I think:

> `annotations` is an array of _annotation segment objects_

reads better than

> `annotation` is an array of _annotation segment objects_

but either way.

kpreid:
FWIW, I have already [said in #35 that I think they should be plural](https://github.com/gnuradio/SigMF/issues/35#issuecomment-303485097). That the examples agree is even better!

bhilburn:
Hey @djanderson - Agreed! Yes, let's make them plural. If you want to submit a PR, I'll get it merged ASAP (I promise! ;) ).

Thanks so much for raising this `=)`

bhilburn:
Fixed by the merge of #59!

Issue #35, opened 2017-04-11
Use of the `segment` name


bhilburn:
In the discussion in #23, there was some feedback regarding the use of the word `segment`:

> The "segments" is a pretty bad name. When discussed live at the hackfest, the goal was never to have disjoint piece of data (you'd use several files for that). (that's my understanding at least).
> 
> They were just written that way because the parameters / infos / metadata written in them are valid until the next segment 'sample_start', rather than have a fixed sample length. Basically to store things like frequency changes or gain changes or things like that in a continuous capture.
> 
> But this whole "segment" naming confuses things more than anything IMHO.

So, the way I think of `segments` is that the datafile itself does not exist in the form of segments, *at all*. It is one monolithic data dump. A `segment` is simply a single `JSON Object` in one of the top-level objects that has some metadata that applies to some portion of the datafile.

What are others thoughts on this? Are we using the term in a confusing way? Is it a bad term to have chosen in the first place? Perhaps we can make this better by more clearly defining `segment` in the spec?

kpreid:
It sounds like you're talking about _capture segments_, but you're just saying “segment”. Therefore I point out the terminology is subject to misuse in a different way than you meant to raise!

bhilburn:
So, I think of `segments` as what we now call `Capture Segment Objects` and `Annotation Segment Objects`.

So, for example, this is a capture segment:
```
        {
            "core:sample_start": 100000,    # Mandatory
            "core:sample_rate": 10000000,   # 10 MHz
            "core:frequency": 950000000,    # Now at 950 MHz
        },
```

And this is an annotation segment:
```
        {
            "core:sample_start": 1000000,   # The sample index at which this annotation first applies.
            "core:sample_count": 120000,    # The number of samples that this annotation applies to.
            "core:comment": "Some text comment about stuff happening",
        },
```

The spec definitely uses the word `segment` in what appears to be an ambiguous way, though, as currently written.

bhilburn:
So, after reviewing the use of the word `segment` in the spec, I actually don't see any examples of it being used inconsistently. The changes from @kpreid in f842a67e actually made this even more clear, I think.

I'll leave this issue for a bit longer to give folks the opportunity to comment or suggest improvements, but in lieu of that I'm considering this a "won't fix".

bhilburn:
One thing we could do to clarify this is specifically describe how we are using the word `segment` in the document. That might be the best answer.

Any thoughts on the best way to go about doing that?

[...]

PR #59, opened 2017-08-16
Use plural key names when value is an array

djanderson:
This PR pluralizes the names of top-level metadata keys if their value is an array.

Specifically, it changes "capture array" to "captures array", and "annotation array" to "annotations array". It leaves references to "capture segment object" and "annotation segment object" as-is.

It also updates the example_metadata.py file to reflect these and previous changes to the spec.

See issues #35 and #56 for discussion.

### DIFF
--- a/example_metadata.py
+++ b/example_metadata.py
@@ -4,25 +4,23 @@
 
 {
     "global": {
-        "core:datatype": "cf32",            # The datatype of the recording (here, complex 32-bit float)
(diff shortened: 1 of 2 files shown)


bhilburn:
Nice work, @djanderson! Thanks so much for taking this on.

In specs, precision is important. I'm glad you spotted this & called it out.

PR #64, opened 2017-09-15
Fix minor typo

fwunsch:


bhilburn:
Thanks so much, @fwunsch!

Especially in specifications, being exact and pedantic matters. Thanks for catching this and filing a PR =)

PR #67, opened 2017-09-18
add file string and repr methods

n-west:
Helps to easily dump the representation in notebooks and repl shell


bhilburn:
Nice! Easy to implement and immediately useful.

Thanks, @n-west!

Issue #65, opened 2017-09-16
Clarify complex sample ordering


storborg:
The spec is currently ambiguous about the ordering of complex samples--that is, IQIQIQ... or QIQIQI.... Here some possible solutions:

1. **State that all complex samples should be IQ.** This makes reader implementation a bit simpler, and keeps the spec smaller, but probably makes some writers (especially converters) a bit slower. I suspect that most people implementing have just been assuming this is the case anyways.
2. **Change the ``c`` designator to either be ``iq`` or ``qi``.** For example, complex 32-bit little-endian floats in IQ order would be ``iqf32_le``.
3. **Add a new ``ordering`` key which could specify either.** This also opens up potential for alternative weird packing options to the spec in the future, like ``iiiiqqqq``, although I can't imagine a use case for such things which would justify the hassle.

Personally, I favor option 1, as I don't think it's too computationally expensive to swap QI for IQ.

kpreid:
Avoid unnecessary degrees of freedom. Option 1.

bhilburn:
Thanks so much for raising this both at GRCon and now here, @storborg!

I agree with both you and @kpreid - I think Option 1 is the way to go. 

@storborg - Are you at all interested in making a PR for this? If not, I can take a crack at it in a bit.

storborg:
@bhilburn Sorry, I've been delinquent on this but I'll take a crack at a PR right now.

bhilburn:
Fixed by #77. Thanks, @storborg!

PR #77, opened 2017-10-02
Clarifying the order of complex samples per #65.

storborg:
Adds a note to the specification document clarifying IQ vs QI ordering.

I don't think this spec change affects validation in any way, so this should be all that's required.


bhilburn:
Thanks so much, @storborg! Much appreciated `=)`

Issue #61, opened 2017-09-12
Incorrect tar format in python code


pwicks86:
The python code creating sigmf archives in this repository uses the wrong format for it's sigmf files. The spec specifies that archives must be in  POSIX.1-2001 format. By default, the python tarfile module creates GNU tar files, not  POSIX.1-2001. [Line 78 of archive.py](https://github.com/gnuradio/SigMF/blob/master/sigmf/archive.py#L78) should be `sigmf_archive = tarfile.TarFile(mode="w", fileobj=sigmf_fileobj, format=tarfile.PAX_FORMAT)`

I can provide a PR for this as well, but it's a pretty tiny change

djanderson:
@bhilburn, feel free to assign me. I'll add a test for the format as well.

bhilburn:
Excellent catch, @pwicks86! Thanks so much for spotting this and then raising it. Especially in standards / specifications, being pedantic to the extreme is the best way to guarantee that everyone walks away with the same understanding of the spec.

@djanderson - Sounds great! Go ahead and submit the PR whenever you're ready `=)`

mbr0wn:
@bhilburn I've invited @djanderson to gr-devs, when he accepts, you can assign to him.

bhilburn:
Fixed by #78. Thanks, everyone!

PR #78, opened 2017-10-06
Read and write POSIX.1-2001 format tar archives

djanderson:
Thanks to @pwicks86 for catching the discrepancy between the spec and the implementation!

Closes issue #61.


bhilburn:
Thanks so much, @djanderson and @pwicks86!

PR #86, opened 2017-11-29
First shot at allowing for extension namespaces.

bhilburn:
Addressing the long-running discussion in #13, this PR is my first take on describing how to create extension namespaces and defining the concept of canonical namespaces.

Please review and comment!

Issue #75, opened 2017-09-20
Canonical Extension Namespace: Modulation / Signal Type


bhilburn:
One of the most popular annotation types is going to be signal / modulation type. Per #13, for anything that we know is going to be used heavily, we are going to create canonical extension namespaces so that we don't end up with a bunch of competing sub-standards for describing annotations in SigMF.

(Recall, supporting canonical extension namespaces are *not* required to be SigMF `compliant`, but simply provide a common namespace for those that want it).

Thus, I would like to start hammering down how this namespace works. In my mind, there is a taxonomy of signal types that people will want to use. For example:

* The system is using 64QAM
* The system is using OFDM with 64QAM
* The system is using 802.* with OFDM-64QAM
* The system is a DJI drone signal link

So, a few questions based on this:

1. How far should we go?
2. Am I missing a layer that makes sense to add?
3. How do we represent these?

kpreid:
Thoughts:

* I believe there are not any already-existant vocabularies for describing _all_ of these layers. You will have to invent some of them, and some of them need to be fairly freely extensible (e.g. "DJI drone signal link" as you said is just one company's invention and there will be many, many more like that).

* Arguably, you should not have a fixed number of layers. There is always the possibility of someone adding another protocol on top of an existing one (though they may become more and more irrelevant to RF signal analysis), or at the other end sending a perfectly good signal into a FM modulator.

mbr0wn:
This is an extremely difficult problem. There's such a large number of combinations. Professional taxonomers are there for a reason...
Based on, in particular, SIGINT efforts I've seen here, I would expect that whatever we do here will fail at some point. We could also just not try, and make the modulation / signal type a free text field.

bhilburn:
My problem with making it a free-text field is that we will end up creating many "standards within the standard". I really think we need a canonical way of representing this, but it needs to be flexible, per @kpreid's comments.

I wonder if there are other efforts that have solved similar problems from which we could draw?

bhilburn:
So, what are people's thoughts on defining a canonical listing for "common" modulations / PHYs, a way of specifying a hierarchy, and allowing free-form definition of everything else?

It's a compromise, and is open to misuse, but also provides some degree of portability. Thoughts?

kpreid:
[...]

PR #87, opened 2017-11-29
Extension Namespace: Modulation

bhilburn:
Okay, so this is a **WIP** branch for #75, creating a canonical extension namespace for describing communications systems, including modulation.

There's obviously more that needs to be done even for the first draft to be complete, but my first pass at mapping @dimitriblock's suggestion to a modulation description is complete.

Additionally, this PR is the first example of a format for a canonical extension namespace.

Note that this branch is built on top of the work done in PR #86.

Please review, comment, and provide feedback.  Just getting this draft took a not insignificant amount of effort, and I wanted to get it pushed publicly so we can start iterating. The goal is to provide obvious ways to describe common modulations while still allowing for the creation of arbitrary new ones.

### DIFF
--- a/extensions/modulation.sigmf-ext.md
+++ b/extensions/modulation.sigmf-ext.md
@@ -1,4 +1,4 @@
-# The `modulation` SigMF Extension Namespace v0.0.1
+# The `modulation` SigMF Extension Namespace v0.0.2
 
 ## Description
 
@@ -32,95 +32,165 @@ specified.
 
 ## Specification
 
-This specification defines new name/value fields in the `modulation` namespace
+This specification defines an extension `modulation` namespace
 that can be used in SigMF `annotations` to describe wireless communications
 systems.
 
 ### Annotations
 
+This extension adds the following field to the `annotations` global SigMF object:
+
 |name|required|type|description|
 |----|--------------|-------|-----------|
-|modulation|false|string|Describes a communications modulation.|
-|system|false|string|Describes the communications system at a higher-level.|
+|modulation|false|object|Describes a communications system.|
 
-##### The `modulation` Field
+The field of communications is vast, and there may be communications systems that cannot be described using the names and fields described in this extension. If you need additional or different fields to describe a system, create a new extension that adds the necessary fields to the `modulation` namespace and/or submit the new fields to be upstreamed into this specification.
 
-The following initialisms and acronyms are used in the ABNF form of this
-metadata field:
+#### The `modulation` Object
 
-* `analog-type`:
-    * **am**: amplitude modulation
-    * **fm**: frequency modulation
-    * **pm**: phase modulation
-    * **ssb**: single side-band
-    * **dsb**: dual side-band
-    * **vsb**: vestigial side-band
-* `analog-carrier-variant`:
-    * **wc**: with-carrier
-    * **sc**: suppressed-carrier
-    * **rc**: reduced-carrier
-* `digital-type`:
-    * **ask**: amplitude-shift keying
-    * **fsk**: frequency-shift keying
-    * **psk**: phase-shift keying
-    * **qam**: quadrature-amplitude modulation
-    * **ook**: on-off keying
-    * **cpm**: continuous phase modulation
-    * **msk**: minimum-shift keying
-* `common-psk`:
-    * **b**: binary
-    * **q**: quadrature
-* `digital-type-variants`:
-    * **d**: differential
-    * **o**: offset
-    * **s**: staggered
-* `digital-shared-type`:
-    * **m**: multiplexing
-    * **ma**: multiple-access
-* `digital-multi-type`:
-    * **fd**: frequency-division
-    * **ofd**: orthogonal frequency-division
-    * **td**: time-division
-    * **pd**: phase-division
-    * **cd**: code-division
-* `spread-spectrum`:
-    * **dsss**: direct-sequence spread spectrum
-    * **css**: chirp spread spectrum
-    * **fhss**: frequency-hopping spread spectrum
+`modulation` objects contain name/value pairs that describe the modulations of communications systems. These objects MAY contain the following pairs:
 
-The `modulation` field must be defined according to these rules:
+|name|required|type|
+|----|--------------|-------|
+|`type`|false|string|
+|`class`|false|string|
+|`carrier_variant`|false|string|
+|`symbol_variant`|false|string|
+|`order`|false|uint|
+|`duplexing`|false|string|
+|`multiplexing`|false|string|
+|`multiple_access`|false|string|
+|`spreading`|false|string|
+|`bandwidth`|false|float|
+|`system`|false|string|
 
-```abnf
-modulation = analog / digital
+###### The `type` Pair
 
-new-type = 1*32(DIGIT / ALPHA)
+The `type` name can have the following values:
 
-analog = analog-type ["-" analog-carrier-variant]
+|value|description|
+|----|-------|
+|`analog`|analog modulation scheme|
+|`digital`|digital modulation scheme|
 
-analog-type = amplitude-modulation / "FM" / "PM" / new-type
-amplitude-modulation = "AM" / "DSB" / "SSB" / "VSB"
-analog-carrier-variant = "WC" / "SC" / "RC"
+###### The `class` Pair
 
-digital = [digital-multi-type "-"] [digital-symbols "-"] digital-type ["-" spread-spectrum]
+The `class` name can have the following values:
 
-digital-multi-type = (("FD / "OFD") / "TD" / "PD" / "CD" / new-type) [digital-shared-type]
-digital-symbols = 1*32(DIGIT)
-digital-shared-type = "M" / "MA" / new-type
+|value|description|
+|----|-------|
+|`am`|(analog) amplitude modulation|
+|`fm`|(analog) frequency modulation|
+|`pm`|(analog) phase modulation|
+|`ssb`|(analog) single side-band|
+|`dsb`|(analog) dual side-band|
+|`vsb`|(analog) vestigal side-band|
+|`ask`|(analog) amplitude-shift keying|
+|`fsk`|(digital) frequency-shift keying|
+|`psk`|(digital) phase-shift keying|
+|`qam`|(digital) quadrature-amplitude modulation|
+|`ook`|(digital) on-off keying|
+|`cpm`|(digital) continuous phase modulation|
+|`msk`|(digital) minimum-shift keying|
 
-digital-type = [digital-type-variant] ("QAM" / "ASK" / "FSK" / "PSK" / "OOK" / "CPM" / "MSK" / new-type)
-digital-type-variant = "D" / "O" / "S" / new-type
+###### The `carrier_variant` pair
 
-spread-spectrum = "DSSS" / "CSS" / "FHSS" / new-type
-```
+The `carrier_variant` name can have the following values:
 
-Note: The modulation strings "BPSK" and "QPSK" are _not_ supported by this
-syntax. The canonical way to specify those modulations is "2-PSK" and "4-PSK",
-respectively. One of the primary goals of this specification is to create
-strings that are easily parsed and understood by computers, and allowing for
-the short-hand version of those strings adds complexity to that processing
-logic, especially when combined with other modifiers like "differential".
+|value|type|description|
+|----|----|-------|
+|`with_carrier`|(analog) with-carrier modulation|
+|`suppressed_carrier`|(analog) suppressed-carrier modulation|
+|`reduced_carrier`|(analog) reduced-carrier modulation|
+|`single_carrier`|(digital) single-carrier modulation|
+|`multi_carrier`|(digital) multi-carrier modulation|
 
-##### The `system` Field
+###### The `carrier_variant` pair
 
-TODO 
-describes `802.11ac` or `Wireless Webcam`
+The `symbol_variant` name can have the following values:
+
+|value|type|description|
+|----|----|-------|
+|`differential`|differential modulation|
+|`offset`|offset modulation (sometimes called 'staggered')|
+
+###### The `order` Pair
+
+The `order` name has an unsigned integer value that describes the modulation order, which typically refers to the number of symbols or states in a digital modulation (e.g., QAM64 has 64 symbols, QPSK has 4 symbols).
+
+###### The `duplexing` Pair
+
+The `duplexing` name can have the following values:
+
+|value|description|
+|----|-------|
+|`tdd`|time-division duplexing|
+|`fdd`|frequency-division duplexing|
+
+###### The `multiplexing` Pair
+
+The `multiplexing` name can have the following values:
+
+|value|description|
+|----|-------|
+|`tdm`|time-division multiplexing|
+|`fdm`|frequency-division multiplexing|
+|`cdm`|code-division multiplexing|
+|`ofdm`|orthogonal frequency-division multiplexing|
+|`sdm`|space-division multiplexing|
+|`pdm`|polarization-division multiplexing|
+
+###### The `multiple_access` Pair
+
+The `multiple_access` name can have the following values:
+
+|value|description|
+|----|-------|
+|`fdma`|frequency-division multiple access|
+|`ofdma`|orthogonal frequency-division multiple access|
+|`tdma`|time-division multiple access|
+|`cdma`|code-division multiple access|
+|`sdma`|space-division multiple access|
+|`pdma`|power-division multiple access|
+
+###### The `spreading` Pair
+
+The `spreading` name can have the following values:
+
+|value|description|
+|----|-------|
+|`fhss`|frequency-hopping spread spectrum|
+|`thss`|time-hopping spread spectrum|
+|`dsss`|direct-sequence spread spectrum|
+|`css`|chirp spread spectrum|
+
+###### The `bandwidth` Pair
+
+The `bandwidth` name has a float value that describes the channel bandwidth of the signal. Note that this is different from what is reported in the `core` namespace within an annotation, describing the occupied spectrum of a signal, which may or may not be comparable to the actual channel bandwidth of the communications system.
+
+###### The `system` Pair
+
+The `system` name is meant to provide a way to describe the higher-level system. For example, "802.11ac", "BlueTooth", or "LTE Release 12".
+
+## Examples
+
+Here is an example of a relatively simple modulation label, which describes a 10 kHz FM signal using time-division duplexing:
+
+    "modulation": {
+        "type": "analog",
+        "class": "fm",
+        "duplexing": "tdd",
+        "bandwidth": 10000.0
+    }
+
+Here is a more complex example that describes an LTE 5 MHz SC-OFDMA downlink:
+
+    "modulation": {
+        "type": "digital",
+        "class": "qam",
+        "carrier_variant": "single_carrier",
+        "order": 16,
+        "multiple_access": "ofdma",
+        "bandwidth": 5000000.0,
+        "system": "LTE Release 12"
+    }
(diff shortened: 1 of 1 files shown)


bhilburn:
I just pushed a change to this which fixed two of the biggest problems I had with the first draft, which was how the `symbol-count` was defined, and how to deal with `BPSK` and `QPSK`, which don't follow the normal rules of modulation strings.

bhilburn:
I wanted to share one of the major items I don't like about the current design:

The goal is to reduce the ambiguity of parsing a particular modulation to make these easily consumable by a computer. At the same time, though, it would be impossible to specifically call out every possible "accepted" modulation - which is why the arbitrary string `new-type` exists.

We then suffer from the problem, then, of being able to use `new-type` to recreate modulation types that are already defined by the spec.

For example, following the spec properly, to indicate `differential QPSK`, you would write `4-DPSK`. But, if you abuse the `new-type` string, you could just as easily write `DQPSK` - which, while accepted in common usage, is not canonical to the spec.

I'm not sure how to solve this problem. Any thoughts / input others have would be greatly appreciated!

bhilburn:
Okay, I took a second crack at this, which completely reformats the extension. Instead of a single ABNF string, it is now a JSON object.

The ABNF string had some nice qualities, principally that there was no ambiguity in the types of systems you could define (e.g., you couldn't define something with an analog modulation and a symbol order) and that it was an *ordered* structure, which simplifies some aspects of the parsing logic.

But, at the same time, parsing the string was more complex than I would like for it to be, and if your application only cared about some of the fields in the string you still had to traverse other fields.

This new model is more flexible, which in some ways is bad (you can create descriptions that are non-sensical, mixing analog and digital aspects), but is simpler to create, easily filterable, and easy to reference / parse by applications.

I'm super interested in feedback, here, both on the design and on the fields in the extension!

MichaelCotton:
We have a similar emerging SigMF namespace that we call `sigmf-ns-waveform` (currently private). In summary: 
- the purpose of sigmf-ns-waveform is to provide models and parameters for textbook and standardized complex-baseband waveforms. The intention is to build a library of simulated waveforms for testing and training signal identification algorithms. The waveform library can also be used for signal generation purposes in system level interference tests.
- See `waveform` in the sigmf-ns-scos `Transmitter` object at
 https://github.com/NTIA/sigmf-ns-scos#32-transmitter-object
- We don't have the ambiguity problem that you refer to in your comment above bc the models are based on actual waveforms, e.g., IEEE 802.11p.
- The `Transmitter` object can be used to specify a known transmitter that we are looking to monitor - see `transmitter_definition` at
https://github.com/NTIA/sigmf-ns-scos#3-global
- Similar to your purposes, the `Transmitter` object can be used for signal identification - see `transmitter_identification` at
https://github.com/NTIA/sigmf-ns-scos#5-annotations

bhilburn:
@MichaelCotton - This is really interesting and helpful! Thanks so much for commenting and providing links. I'll for sure be interested in seeing your `waveform` namespace if/when it becomes public.

I really like the concept of the `transmitter` object. It's a very clean way to describe an emitter. One of the more interesting design decisions, here, is making the `waveform` (i.e., the signal) a property of the `transmitter` rather than the other way around. I can see good reasons to do both.

I need to spin on whether we should have something like the `transmitter` object as a canonical namespace. We for sure need to cover the functionality, but I'm not sure how best to handle it in a generic fashion.

bhilburn:
A note about the new design now in this PR that I pushed earlier:

One solution to the ambiguity problem I mentioned (where you *could* define nonsensical annotations, like digital fields for analog modulations) would be for `analog` and `digital` to be objects themselves. For example:

    "modulation": {
        "analog": {
            "class": "fm",
            "duplexing": "tdd"
        },
        "bandwidth": 10000.0
    }

    "modulation": {
        "digital": {
            "class": "qam",
            "carrier_variant": "single_carrier",
            "order": 16,
            "multiple_access": "ofdma"
        },
        "bandwidth": 5000000.0,
        "system": "LTE Release 12"
    }

I didn't do this originally, because it introduces some complexities, like:

1. The same name, say, for example, `class`, has different allowed values depending on whether or not it's part of an `analog` object or a `digital` object.
2. One way to get around #1 would be to have different name pairs for `analog` vs `digital` (e.g., `digital-class`), but then you can't filter on pairs in the sub-object without conditionalizing the parent.

Right now, I'm thinking those compromises might be worth it, though, for clarity and cleanliness. I'm favoring #1, at the moment. Will keep spinning, though, and am open to feedback!

bhilburn:
So, after spinning on this, I think the structure as it already exists in the PR is probably best (i.e., a single object, **not** nested objects as proposed in my comment, above).

My reasoning, here, is:

1. Parsing and filtering over the nested object is more difficult.
2. I don't want to preclude cases we haven't thought of (e.g., someone incorporates an analog component into a typically digital modulation).
3. I really don't like that it effectively creates two sets of possible values for fields like `class`.

The reason for not using this design is that it made it "makes it possible to create modrec annotations that don't make sense" (mixing analog and digital features). While that's true, the nested structure doesn't totally solve that problem, and actually introduces new ones (like the third item in the list above). I think the increased ease-of-use and flexibility win out, here, especially since the more complex design doesn't completely solve the problem.

Last item, here, is that the name `modulation` is a bit of a misnomer, as this encompasses more than just modulation. Right now, I'm thinking about renaming it to`communications`.

Happy to hear feedback on all of the above, and otherwise will move ahead with the current design & suggested rename shortly!

extensions/modulation.sigmf-ext.md
```diff
@@ -0,0 +1,196 @@
...
+|`fsk`|(digital) frequency-shift keying|
+|`psk`|(digital) phase-shift keying|
+|`qam`|(digital) quadrature-amplitude modulation|
+|`ook`|(digital) on-off keying|
+|`cpm`|(digital) continuous phase modulation|
+|`msk`|(digital) minimum-shift keying|
+
+###### The `carrier_variant` pair
+
+The `carrier_variant` name can have the following values:
```
n-west: Unless someone with the need for this to be described separately this is either overly complex and incomplete or part of the modulation. For example, is there anything other than AM that would use a suppressed carrier? In that case we should just have a modulation type that is AM-SC. In the multi-carrier (OFDM) case the `class` should be OFDM because none of the other classes make sense.

extensions/modulation.sigmf-ext.md
```diff
@@ -0,0 +1,196 @@
...
+|----|----|-------|
+|`with_carrier`|(analog) with-carrier modulation|
+|`suppressed_carrier`|(analog) suppressed-carrier modulation|
+|`reduced_carrier`|(analog) reduced-carrier modulation|
+|`single_carrier`|(digital) single-carrier modulation|
+|`multi_carrier`|(digital) multi-carrier modulation|
+
+###### The `carrier_variant` pair
+
+The `symbol_variant` name can have the following values:
```
n-west: The section heading should probably be `symbol_variant`, but this is another key that I'm not sure makes 100% sense. For the offset case, this is just another modulation, (it's OQPSK or it's QPSK). For the differential case, to be fair you can't identify that at the PHY, so maybe it's just a true or false? It's hard to say since I don't have a practical need for declaring if a signal is differential or not, but I don't think a single value should declare differential OR offset (because they could be both)

extensions/modulation.sigmf-ext.md
```diff
@@ -0,0 +1,196 @@
...
+The `spreading` name can have the following values:
+
+|value|description|
+|----|-------|
+|`fhss`|frequency-hopping spread spectrum|
+|`thss`|time-hopping spread spectrum|
+|`dsss`|direct-sequence spread spectrum|
+|`css`|chirp spread spectrum|
+
+###### The `bandwidth` Pair
```
n-west: this seems duplicated by the existing upper and lower freq_edges described in the annotation field.

extensions/modulation.sigmf-ext.md
```diff
@@ -0,0 +1,196 @@
...
+Here is an example of a relatively simple modulation label, which describes a 10 kHz FM signal using time-division duplexing:
+
+    "modulation": {
+        "type": "analog",
+        "class": "fm",
+        "duplexing": "tdd",
+        "bandwidth": 10000.0
+    }
+
+Here is a more complex example that describes an LTE 5 MHz SC-OFDMA downlink:
```
n-west: an OFDMA signal can have multiple `class`es in it, which is why I think OFDM should be placed in to the `class` object and possibly a new OFDM  object should be created which should be described by someone who has a need to describe detailed descriptions of OFDM signals.

extensions/modulation.sigmf-ext.md
```diff
@@ -0,0 +1,196 @@
...
+|value|type|description|
+|----|----|-------|
+|`differential`|differential modulation|
+|`offset`|offset modulation (sometimes called 'staggered')|
+
+###### The `order` Pair
+
+The `order` name has an unsigned integer value that describes the modulation order, which typically refers to the number of symbols or states in a digital modulation (e.g., QAM64 has 64 symbols, QPSK has 4 symbols).
+
+###### The `duplexing` Pair
```
n-west: I think this should be scrapped unless someone with a need can explain why it's necessary. I would probably want to describe each signal as its own annotation in the frequency-duplexing case and in the time-duplexing case describe each time slot with its own annotation.

n-west:
I think this should be simplified to the core components of a modulation: the order, class, and maybe spreading to keep it simple for the first revision. If more complex fields are needed then we need input from real users who are developing systems using those fields or we may come to a solution that doesn't handle the real needs. The system description is good and should be outside of the modulation.

extensions/modulation.sigmf-ext.md
bhilburn: I think this is a really good point. Removing.

bhilburn:
Okay, after reading your comments and thinking more, I want to move on to a third iteration of this design. Basically, my thinking is that we get to a point where it's easy to assign labels for the simple use cases, and we provide a mechanism that can be used to describe more complex systems without actually trying to specify the entirety of those fields, which just isn't realistic.

What that looks like, I'm still spinning on... will push a draft for review with some new ideas once I land on something.

bhilburn:
Okay, I realize it's been, uh, 8 months, but I hadn't forgot about this.

So, first, a few high-level comments:

- Per my previous comment, the name `modulation` doesn't really make sense as an extension, as the goal is to be able to describe far more than just the modulation of a signal.
- This extension should provide a consistent mechanism to be understood without further context about the "label" of a signal.
- IF the writer application wants to add additional metadata regarding that label, it should be able to.

Given these, here is an example of my proposed structure:

```json
{
    ...<other top-level objects here>
    ...,
    "annotations": [
        {
            "core:sample_start": 0,
            "core:sample_count": 1000000,
            "modulation:signal": {
                "label":"wifi",
                "wifi": {
                    "standard":"802.11ac",
                    "modulation":"64-qam",
                    "channel": 8
                },
                "emitter":{
                    "seid": 1,
                    "manufacturer":"linksys",
                    "geo":<wgs84>
                }
            }
        }
        ... <more annotations here>
    }
}
```

So, note a few things:

- Yes, I'm still using the name `modulation` in the example above, but that will change. I'm thinking `signalrec`?
- The string used as the value for the `label` can be used as a key to pull out additional info. It is up to the user to define what goes under that label with another extension spec. We can maybe provide some recommendations for common ones, like what to use for "modulation", for example.
- So the things that this extension MUST define are the keys `label` and `emitter`.

So, questions:

1. Thoughts on the structure, generally?
2. Are there keys other than `label` and `emitter` that should be added to the canonical extension spec for the `modulation:signal` object?
3. What keys should be under `emitter`?

MichaelCotton:
My thoughts on the structure, generally...

From high- to low-level from my perspective...

NTIA will release an `sigmf-ns-ntia` namespace in the fall. The namespace will contain the following extensions: 

- `ntia-core` adds generally useful metadata fields
- `ntia-sensor` defines hardware components and settings
- `ntia-algorithm` describes the measurement performed (detectors, algorithms, etc)
- `ntia-calibration` provides information about calibration factors applied to the data
- `ntia-emitter` gives information about the emitter being measured
- `ntia-waveform` provides metadata to describe measured or transmitted waveforms
- `ntia-environment` gives information about the environment around a sensor or emitter
- `ntia-scos` provides metadata for the NTIA Spectrum Characterization and Occupancy Sensing (SCOS) implementation

For the purpose of defining a targeted and known emitters (e.g., to justify a detection scheme), `ntia-emitter` extends the [Global](https://github.com/gnuradio/SigMF/blob/master/sigmf-spec.md#global-object) with the following name/value pairs:

|name|required|type|unit|description|
|----|--------------|-------|-------|-----------|
`emitters`|false|Array of [Emitter](#11-emitter-object)|N/A|Metadata that describe emitters

where the `Emitter` object has the following properties:

|name|required|type|unit|description|
|----|--------------|-------|-------|-----------|
|`id`|true|string|N/A|Unique id of the emitter.|
|`description`|false|string|N/A|Description of the emitter.|
|`power`|false|double|dBm|Power referenced to antenna input.|
|`antenna`|false|[Antenna](ntia-core.sigmf-ext.md#antenna)|N/A|Metadata that describes the antenna.|
|`transmitter`|false|Transmitter|N/A|Metadata that describes the transmitter.|
|`waveform`|false| [Waveform](ntia-waveform.sigmf-ext.md)|N/A|Metadata that describes transmitted waveform.|
|`latitude`|false|double|decimal degrees|Latitude.|
|`longitude`|false|double|decimal degrees|Longitude.|
|`altitude`|false|double|meters|Height above mean sea level.|
|`speed`|false|double|m/s|Speed.|
|`bearing`|false|double|degrees|Direction (angle relative to true North).|

For sigid and geolocation purposes, `ntia-emitter` defines the `EmitterAnnotation` segment that extends `ntia-core`. `EmitterAnnotation` has the following properties:  

|name|required|type|unit|description|
|----|--------------|-------|-------|-----------|
|`id`|true|string|N/A|Unique id of the emitter.|
|`power`|false|double|dBm|Power referenced to antenna input.|
|`antenna`|false|[Antenna](ntia-core.sigmf-ext.md#antenna)|N/A|Metadata that describes the emitter's antenna|
|`waveform`|false| [Waveform](ntia-waveform.sigmf-ext.md)|N/A|Metadata that describes transmitted waveform.
|`altitude`|false|double|meters|Height above mean sea level.|
|`speed`|false|double|m/s|Speed.|
|`bearing`|false|double|degrees|Direction (angle relative to true North).|

**Structurally, note that `waveform` is part of the emitters object.** For example, 

```json
{
  "global": {
    ...
  },
  "captures": [
    ...
  ],
  "annotations": [
    {
      "ntia-core:annotation_type": "EmitterAnnotation",
      "core:sample_start": 0,
      "core:sample_count": 1024,
      "core:latitude": 40.5,
      "core:longitude": -105.7,
      "ntia-emitter:id": "emitter_1"
      "ntia-emitter:waveform": {
        ...
      },
    },
    {
      "ntia-core:annotation_type": "EmitterAnnotation",
      "core:sample_start": 1024,
      "core:sample_count": 1024,
      "core:latitude": 41.5,
      "core:longitude": -105.3,
      "ntia-emitter:id": "emitter_1"
      "ntia-emitter:waveform": {
        ...
      },
    }
  ]
}

}
```

`ntia-waveform`  will contain a library of signals with parameters and attributes. For example, `IEEE80211p` extends `waveform` with the following additional properties:

|name|required|type|unit|description|
|----|--------------|-------|-------|-----------|
|`info_bit_generation`|false|string|N/A|Model that defines information bit generation. E.g. `"PN"`.|
|`coding_rate`|false|array [k, n]|bits|For every k bits of useful information the coder generates n bits of data. E.g. [1, 2], [2, 3], [3, 4].|
|`packet_length`|false|integer|bits|Packet length.|
|`modulation`|false|string|N/A|Modulation, e.g., `"BPSK"`, `"QPSK"`, `"16QAM"`, `"64QAM"`.|
|`encoder`|false|string|N/A|Description of encoder. E.g. `"Convolutional"`.|
|`number_of_subcarriers`|false|integer|N/A|Number of subcarriers.|
|`number_of_data_subcarriers`|false|integer|N/A|Number of data subcarriers.|
|`number_of_pilots`|false|integer|N/A|Number of pilots.|
|`cyclic_prefix`|false|integer|bits|Size of cyclic prefix.|
|`short_inter_frame_space`|false|float|microseconds|Time required to process a received frame and to respond with a response frame.|
|`preamble_frame`|false|array|N/A|Preamble of 0's and 1's used for synchronization and id beginning of frame.|
|`number_of_info_bits`|false|integer|N/A|Number of information bits.|
|`signal_to_noise_ratio`|false|float|dB|Signal-to-noise ratio. If unspecified, assumed no noise present.|

This is one of many well-defined waveform models to be offered. I recommend we use comm theory as a basis for waveform models.

Some parameters can be adopted at a higher level, for example an IEEE802 level. That will take a little time to flesh out.

There are general measured signal attributes that can be offered, e.g., `measured-3dB-bandwidth`, but that falls within `ntia-algorithm` for us.

bhilburn:
This extension will be complete and merged for v0.0.2

@MichaelCotton - thanks for your comments! Will circle back around and respond soon.

bhilburn:
@MichaelCotton - Thanks so much for posting so much detail about NTIA's upcoming implementation. This was tremendously useful, and your insight is really valuable.

Structurally, it's interesting that my draft proposal in my previous comment, and what you have implemented, are slightly different. In my suggestion, details about the `emitter` and the `waveform` are within the overall `signal` annotation (as in, it's an object within the object). In yours, it's actually structured through the namespace - there is an `emitter` namespace, of which `waveform` is a part.

The list of fields you are including in your namespaces is also really insightful and useful. I like how you've broken down common waveform design properties into generic comms system descriptions.

After reading through your post several times and spinning on this a bit, here is what I am thinking:

1. Definitely using the term `waveform`, hah.
2. I'm leaning towards something of a combination of our designs structurally. I like having it all contained in one object at the annotation-level, because that makes it pretty easy to parse out specifically that object with something like `json-path` for processing. But, internally having them namespaced as you do, as that also better enables parsing, and also makes it simpler for additional extensions to add their own fields to those objects - indeed, it would actually make it possible for someone to use the NTIA's waveform descriptions within this structure.
3. I'm changing the previous design, where the 'value' of the `label` field was the same as the name of the object that contained additional information, such that the latter is always named `details`, regardless of the label. This is far more portable, I think.

So, a modified version of the example I gave previously with the changes discussed above:

{
    ...<other top-level objects here>
    ...,
    "annotations": [
        {
            "core:sample_start": 0,
            "core:sample_count": 1000000,
            "waveforms:signal": {
                "waveforms:label":"wifi",
                "waveforms:detail": {
                    "standard":"802.11ac",
                    "modulation":"64_qam",
                    "channel": 8
                },
                "waveforms:emitter":{
                    "seid": 1,
                    "manufacturer":"linksys",
                    "geo":<wgs84>
                }
            }
        },
        ... <more annotations here>
    ]
}```

The final detail I'm trying to work out, I think, is whether or not to namespace the fields within the sub-objects. So, for example, should `standard` be namespaced within the `waveforms:detail` sub-object? In the NTIA design, each of these objects has it's own namespace. In #105, though, both @n-west and @djanderson raised good counter-points about namespacing at this level.

bhilburn:
Well it's been quite a few months, but I'm circling back around to all of these with the goal of knocking them out.

Here's what I'm thinking in terms of the modulation extension, from the view of an example annotation segment:

```json
{
    "core:sample_start": 0,
    "core:sample_count": 1000000,
    "waveforms:signal": {
        "label": "wifi",
        "detail": {
            "standard": "802.11ac",
            "modulation": "64-qam",
            "channel": 8
        },
        "emitter": {
            "seid": 1,
            "manufacturer": "linksys",
            "geo": "lat/long"
        }
    }
}
```

Key points:

1. Renaming this namespace to "waveforms"
2. The top-level object in an annotation object is namespaced, but the fields within it are not. This isn't actually specific to this extension, but rather something that needs to be standardized within SigMF more largely.
3. The sub-objects within the top-level "signal" are modeled on the structures that NTIA defined in their extension.

I'm going to re-factor this PR, defining those sub-fields per the above, and will update here when it's ready for review.

Atheatos:
Have you had a chance to complete your refactor? I'm looking into converting a dataset to use SigMF and I'm interested in using those fields.

Atheatos:
I made an attempt at the refactor myself based on what I could discern from the discussions above. Let me know what you think.

bhilburn:
@Atheatos - Excellent. Will review ASAP :)

bhilburn:
Closing in favor of focusing discussion in #120

PR #94, opened 2018-04-17
Fix inconsistent type naming

pwicks86:
Everywhere else in the spec uses `uint`, not `uint64`.


bhilburn:
Thanks for the quick fix, @pwicks86!

Issue #81, opened 2017-11-13
Extension Namespace: `antenna`


dharasty:
Is there prior art for a standard way to describe an "antenna pattern"?  By that I mean either a 2D or 3D representation of gain along the antenna's nominal "beam", AND the direction that the beam is "pointed".  (Note that this pattern may be parametric in frequency, as well.)

I'd like to propose that SigMF would be enhanced if it could capture this information in the metadata.

I can consider four subtending use cases:
1) Fixed position / fixed pointing.
    -- example: broadcast antenna.
2) Time-varying position / fixed pointing (or omni antenna).
    -- example: mobile telephony.
3) Fixed position / time-varying pointing.
    -- example: radar.
4) Time-varying position / time-varying pointing.
    -- example: certain exotic military applications.

Let's discuss the first item -- fixed position antenna / fixed pointing -- first.   (If we can't solve that -- or have no interest -- then then al the time-varying cases are moot...)

bhilburn:
Hey @dharasty! No, we have not yet figured out how to describe antennas, yet, but it was identified as an important problem early on (back in #4).

Regarding time-varying stuff: we have decided on a way of handling this in #55 (it's just waiting on someone [*ahem* me *ahem*] to find the time to write it up). So, if we figure out a way to describe antennas, we should be able to just carry it over assuming we design it properly such that it's compatible with our plans in #55.

So, it seems to me that there are three separate items, here:

1. The antenna pattern
2. The antenna position
3. The antenna azimuth

For (2) and (3), we can follow our plan for describing fields that can be fixed or volatile, per #55. Describing (2) should just be the same geographic coordinates that we choose in #73 for the receiver geolocation. I'll leave those discussions to those issues to keep things focused.

So, that leaves two open questions, here, I think:

1. How do we describe antenna pattern?
2. How do we describe antenna azimuth?

Do you have any thoughts on these? Per usual, we don't want to create new standards where standards already exist, so the big question, is, what are good already existing standards for describing these?

bhilburn:
Per my update to #4, this is all going into an extension namespace. I'll take a quick crack at it, but I'm not sure how patterns & azimuths should be described. Suggestions, anyone?

djanderson:
We took a stab at an antenna object in a namespace extension here: https://github.com/NTIA/sigmf-ns-scos#antenna-object

Definitely interested in hearing other ideas as well.

MichaelCotton:
[...]

PR #96, opened 2018-07-18
Proposal to address issue #81

bradeales:
Antenna objects from NTIA SCOS namespace as a proposed solution to issue #81


extensions/antennas.sigmf-ext.md
```diff
@@ -0,0 +1,36 @@
...
+The `Antenna` object contains the following name/value pairs:
+
+|name|required|type|unit|description|
+|----|--------------|-------|-------|-----------|
+|`model`|true|string|N/A|Antenna make and model number. E.g. `"ARA CSB-16"`, `"L-com HG3512UP-NF"`.|
+|`type`|false|string|N/A|Antenna type. E.g. `"dipole"`, `"biconical"`, `"monopole"`, `"conical monopole"`.|
+|`low_frequency`|false|float|Hz|Low frequency of operational range.|
+|`high_frequency`|false|float|Hz|High frequency of operational range.|
+|`gain`|false|float|dBi|Antenna gain in direction of maximum radiation or reception.|
+|`horizontal_gain_pattern`|false|array of floats|dBi|Antenna gain pattern in horizontal plane from 0 to 359 degrees in 1 degree steps.|
```
bhilburn: So, the SigMF spec, I think, in its current form, actually doesn't allow for arrays in the value datatype: https://github.com/gnuradio/SigMF/blob/master/sigmf-spec.md#datatypes

But, I think this is a great example of why we should allow that. Further, looking at json.org, technically, all structures can be `values` in the JSON spec, which means the way we have this written currently kind of breaks that spec. I'm thinking we should add `object` and `array` to the table, which will fix the problem. I can add that change and push it to this branch when you are done making your changes.

extensions/antennas.sigmf-ext.md
```diff
@@ -0,0 +1,36 @@
...
+|`vertical_beam_width`|false|float|degrees|Vertical 3-dB beamwidth.|
+|`cross_polar_discrimination`|false|float|N/A|Cross-polarization discrimination.|
+|`voltage_standing_wave_ratio`|false|float|volts|Voltage standing wave ratio.|
+|`cable_loss`|false|float|dB|Cable loss for cable connecting antenna and preselector.|
+|`steerable`|false|boolean|N/A|Defines if the antenna is steerable or not.|
+|`mobile`|false|boolean|N/A|Defines if the antenna is mobile or not.|
+
+## Annotations
+
+### DynamicAntennaSettings Object
```
bhilburn: I'm interested in your reasoning for creating a new namespace, `DynamicAntennaSettings`, for the fields that go in annotations. I think it would be simpler for them all to be `antenna`, kind of like how we have `core` pairs for all objects.

bradeales:
Tagging @djanderson and @MichaelCotton to include in discussion.

extensions/antennas.sigmf-ext.md
```diff
@@ -0,0 +1,36 @@
+# Antenna Extension
```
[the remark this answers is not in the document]
MichaelCotton: See `version` at https://github.com/NTIA/sigmf-ns-scos#3-global

extensions/antennas.sigmf-ext.md
```diff
@@ -0,0 +1,36 @@
...
+
+The `antenna` namespace extension defines one object for static antenna parameters extending `global` and one object for dynamic antenna parameters extending `annotations`.
```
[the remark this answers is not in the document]
bradeales: I agree. Singular makes sense.

extensions/antennas.sigmf-ext.md
djanderson: You bring up a great point, we pulled this out of our own namespace extension but didn't properly format it as its own standalone namespace. We'll do a bit of refactoring and update this PR.

bhilburn:
Thanks so much for upstreaming this extension, @bradeales @MichaelCotton @djanderson, and for iterating on it for merge, with me! I have merged this into `master`, and it will be included in the next SigMF release `=)`

I versioned it v0.9.0, which gives us some latitude if we need to make changes before calling it "stable".

PR #97, opened 2018-10-01
Removing `length` field from `captures`.

bhilburn:
Per the discussion at GRCon, this field is unnecessary, especially given
SigMF's focus on being a storage format as opposed to a transport
format.

@storborg @pwicks86 - This is a simple change, but this look good to you?


storborg:
Looks good to me.

Issue #69, opened 2017-09-20
Continously Variable Sample Rate


bhilburn:
From GRCon17 discussions:

We already have an effort in-motion to allow for continuously variable fields in #55. Early on in SigMF, though, we wrote into the spec that sample rate was immutable for a recording, and cannot change (see #28 for that discussion).

A couple of people said that this precluded the use of SigMF for them, because they had continuously variable sample rates in their systems (I believe for Doppler compensation). Using what we are doing in #55, applied to `sample_rate` would solve that problem easily.

I think a major question is "can we do this without making reader applications extremely complicated", and I'm not sure what the answer is. If the answer is "no", then perhaps the better solution is to allow this via an extension namespace but not put it in `core` (and thus it wouldn't be required for compliance, but is possible within the standard).

Thoughts and opinions on this?

mbr0wn:
If we come up with a good, generic way to make fields either constant (and then they're easy to modify, and human-readable in the JSON file), or alternatively to be specified separately as another dataset (for continuously changing data), then I don't see why this won't also work in the same way.

Does this make things complicated? That depends. Some applications, where sampling rate really matters, might have trouble with this. But they can always choose to not accept data in this format (after all, even people are already clamouring for this feature, I doubt there'll be more than a few % who will actually use it).
Many applications won't actually care all that much about the sampling rate.

bhilburn:
@mbr0wn - So, I think the main question, here, is whether or not this is something we want to put in `core`. By putting it into `core`. we say that in order for a Reader application (for example) to be compliant, it must be able to parse, interpret, and make use of that field. A continuously variable sample rate, I think, clearly adds complexity to applications that might not want to deal with it; I'm concerned it will unnecessarily raise the burden of writing SigMF-compliant applications.

I *do* want to support this, because it is a legitimate use-case and our work in #55 will make it possible, anyway, but I also want to keep things sane. I think the most obvious solution is to put this in a canonical extension namespace. Thoughts?

bhilburn:
Focusing work & discussion in #99

Issue #68, opened 2017-09-20
Compression for SigMF Archives


bhilburn:
Item from GRCon discussions:

So, we had previously decided in the discussion about archive formats #15 to dis-allow compression, with the reasoning that compressing IQ recordings rarely gives you anything and complicates the reader / writer applications. But, a couple of folks from the National Labs pointed out that you sometimes have recordings of mostly zero values in some systems. They are currently using HDF5 and considering moving over to SigMF in a number of their programs, but would like to see a compression capability, which HDF5 allows for.

I think this argument is reasonable, and it does make sense. There will definitely be recordings where the Vpp is not changing so much as to make compression useless.

I'd like to solicit thoughts and opinions on both this topic, generally, as well as what forms of compression might make the most sense for SigMF archives.

djanderson:
Not having a compression format in the spec doesn't stop people from using compression, though?

Using tar (which we've used as our archive format) as an example, tar doesn't itself specify any compression, but it's almost always compressed, with the compression format tacked on after .tar, so `.tar.gz` or `.tar.bz2`, etc.

Could we recommend a similar syntax, where we use `.sigmf.gz`, `.sigmf.bz2`, `.sigmf.whatever_works_for_you`? 

Pros:
 - simplicity of the spec
 - simplicity of reader/writer applications (they only _have_ to support untarring/tarring, leaving the user to use an external {de}compression utility first, or they could detect the common compression formats and do that for you as well).
 - flexibility to use best compression format for your application

Cons:
 - someone could theoretically compress all their sigmf data with some tool/format that is not free, not OSS, not available on all OS's, etc

kpreid:
This is a side issue to the question of doing it at all but note that extensions like `.sigmf.gz` are bad for use in (at least some) desktop environments because `.gz` would end up matched to a generic decompressor rather than a SigMF viewer which can decompress transparently. This is why when I proposed standardizing an archive format I specified that the extension should be `.sigmf` and not `.tar`.

djanderson:
I'm actually using the SigMF archive format, and I can tell you that 100% of the time I just change the extension to .tar when I download the file so that the built-in archive utility recognizes it and opens it. That's fine, and I get and agree with the point of using the `.sigmf` extension, but also hiding the compression format would just make the file impossible to open with external tools on a system without a dedicated sigmf reader installed.

You can always open the files directly in the sigmf reader that supports compression to avoid it being picked up by a decompression utility first.

[...]

Issue #55, opened 2017-05-18
Volatile Fields


bhilburn:
In #4, the issue of fields that are continuously changing came up:

> So, I think this raises an important question, which is how do we best deal with changing values for any of these fields? What if the antenna is being rotated during the capture to sweep an area (e.g., RADAR). What if the receiver is moving, and so the long/lat fields are changing rapidly?
> 
> Generating a new capture segment for every change would result in a new capture segment for every single sample, which is obviously incredibly cumbersome.

@kpreid threw in the idea of:
> One could add angular velocity fields so that the application can interpolate. That would still require multiple segments to approximate reality, because any given interpolation strategy does not reflect the actual axes of rotation of the antenna rotator or whatever, but it might make the required volume of data more reasonable.

And @patchvonbraun added:
> I'm thinking that there's no clear "best way" to handle these
> situations.  For example, the application could clearly decide for
> itself what granularity to update some types of metadata, which implies
> that "carry over" would be acceptable.
> 
> If one considers, from the radio astronomy perspective, an antenna that
> is tracking a source, it could reasonably want to update az/el
> information on a "routine basis", but certainly not on an
> every-damned-sample basis.
> 
> I wonder if there's guidance on these types of issues from existing
> standards?  Like the FITS standard that is used in science for various
> types of annotated,
>    vaguely image-like, data?

I think this is an important issue to discuss and address to whatever degree we are able.

What if we were to define an ancillary field that defined movement in a generic and standard way in a canonical namespace that could be attached to other fields (like location, antenna, etc.,)?

The hardest part of this, obviously, is "how do you define movement in a standard way" - is anyone aware of an existing standard for this that we could leverage?

Vinz87:
In my opinion, metadata don't change over a dataset, generally speaking.

If the radar's antenna rotates over the acquisition, the antenna pointing is not a metadata anymore, it's another time-varying piece of information, another signal to be sampled, i.e. another dataset.

If the location of the platform changes because it is moving, it stops to be a metadata and it will become another dataset. Also keep in mind that a GPS position is logged at a typically lower rate than the physical measurement of the sensor. Therefore, different rates issues or interpolation needs would come up if thought as a metadata.

kpreid:
That is an interesting angle on the problem and might produce some good ideas.

[...]

Issue #89, opened 2018-01-16
Potentially adding a way to note discontinuities in time to the core namespace for capture segments


pwicks86:
One issue I've run across while building out a set of blocks for SigMF is that it's common to want to create capture segments to note a discontinuity in time of the incoming stream of samples, but not to have a known clock to reference against. For example, if a USRP source has an overflow while receiving samples, it will attach a new `rx_time` tag to the next valid sample to notify downstream users of the discontinuity in data and I'd like my SigMF sink block to be able to handle this. It seems like the `core:datetime` is the place to store this data, but in practice this is difficult to actually achieve, since there isn't a good way to know what the timebase is of the `uhd::time_spec_t` in the `rx_time` tag, it could be unix time or it could be time since the USRP was turned on or some other timebase. Certainly we can handle this via an extension, but it seems like this is a common enough occurrence to warrant inclusion in the core namespace. Off the top of my head, we could add something like this to the core namespace for capture segments:

| name  | required | type | description |
| ------------- | ------------- | ------------- | ------------- |
| core:offset_seconds  | false  | uint64 | The whole seconds part of the offset from the beginning of the recording
| core:offset_frac_seconds  | false | uint64 | The fractional part of the offset from the beginning of the recording

bhilburn:
You raise good points! We had some pretty detailed discussions about time offsets early on, including in this issue, #10, which I recommend reading to get some additional background.

So, as I read your write-up, there are two major aspects, here:

1. Including a time offset field(s) in the `core` namespace: So, fundamentally, I don't think I understand why this is useful - it would be great if you could explain it a bit more for me. My confusion comes down to this point, which I'm copying / pasting over from #10:

> Let's say you have "timestamped" data in the metadata, are visualizing the dataset in a reader application, and want to show the timestamped data. Wouldn't the reader app need to convert the timestamp to a sample index in order to map it in the visualization? If that's the case, why not record it with a sample index? Perhaps this is just a question of where do we put the burden - on the writer or reader. Am I misunderstanding something?

It seems to me that in any case you are using those offset numbers, you are also mapping a time to a sample index, in which case you could just use the sample index and sample rate which is already provided - thus, the offset information is redundant. I'm interested in your take, here.

[...]

PR #99, opened 2018-10-02
Multiple Recordings Extension

bhilburn:
This is a first draft of the multiple recordings functionality that we have been discussing in #55 and #69, and that we discussed heavily in the SigMF workshop at GRCon18.

I renamed the `volatile` canonical extension that we had been discussing to `multirecordings`, which supports linking multiple recordings for both multi-signal captures and multi-channel receivers. The primary mechanism by which this works is simply a new datatype and one new name/value pair in captures. 

Support for the new datatype, `recording`, in other extensions effectively provides the capability to link recordings for that particular field. As a first draft of this, I modified the `antenna` extension to show how this might work.

Because I changing the file structure of the primary specification, I also made the changes for compression discussed in #68. Additionally, I made some formatting changes in the markdown to prettify the tables and update the ToC. I apologize for cluttering this PR with those changes, but at this point I just as soon leave them in, here.

Please review and provide input / feedback / suggestions. This is a significant change that we have been discussing for nearly a year at this point.

Tagging the folks that were actively involved in the discussion at GRCon: @n-west @storborg @pwicks86

### DIFF
--- a/extensions/multirecordings.sigmf-ext.md
+++ b/extensions/multirecordings.sigmf-ext.md
@@ -1,0 +1,103 @@
+# Multi-Recordings Extension v0.0.1
+
+This extension makes it possible to link multiple SigMF recordings together to
+represent a single data-capture event. This is primarily used for two purposes:
+representing aspects of one SigMF recording with another recording, and
+capturing multi-channel data.
+
+For some recordings, there are fields that are continuously variable. Examples
+include the geolocation of a radio receiver that is on a moving vehicle, the
+azimuth of a spinning antenna aperture, and the temperature readings of a
+thermal sensor. If these were stored as metadata in the "primary" signal
+recording, they would bloat and complicate the use of that recording.
+Furthermore, it would be difficult to process any aspect of the data without
+parsing all of the recorded data. This extension enables you to treat such
+fields as *simply another `SigMF Recording`*.
+
+For recordings that digitize multiple streams of information (e.g., a
+phased-array receiver with multiple channels), this extension enables you to
+capture the multiple streams and link them such that each one can be treated as
+a separate `SigMF Recording` or they may be treated together as one.
+
+Per the SigMF specification, all recordings must be independently evaluated and
+be able to stand-alone.
+
+## Datatypes
+
+| type      | long-form name        | description                                             |
+| --------- | --------------------- | ------------------------------------------------------- |
+| recording | string recording name | A string that indicates the name of a `SigMF Recording` |
+
+The `recording` datatype is the primary mechanism by which SigMF Recordings are
+linked. This datatype is the name of a SigMF Recording, **without** extensions
+(i.e., given a recording comprising `N.sigmf-meta` and `N.sigmf-data`, the
+`recording` datatype value would simply be `N`).
+
+The metadata and dataset files of the recording MUST be in the local directory
+(i.e., local to the SigMF Recording in which they are referenced, and thus in
+the same SigMF Archive).
+
+### `core` Namespace Support
+
+If the `multirecordings` extension is supported, it MAY be used in the following
+name/value pairs within the `core` namespace.
+
+#### Capture Segment Objects
+
+The following are existing name/value pairs in capture segment objects in the
+`core` namespace:
+
+| name           | required | type      | description  |
+| -------------- | -------- | --------- | ------------ |
+| `frequency`    | false    | recording | The center frequency of the signal in Hz.|
+
+### `multirecordings` Namespace
+
+This extension adds the following name/value pairs under the `multirecordings`
+namespace.
+
+#### Global Object
+
+| name          | required | type      | description                         |
+| ------------- | -------- | --------- | ----------------------------------- |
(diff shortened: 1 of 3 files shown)


bhilburn:
This must be addressed for v0.0.2 for long-term support & usage

n-west:
Just thinking out loud here of usage scenarios to come back later and identify how they might be handled by this.

1) Multi-antenna with some sort of synchronzed tuning
2) Multi-RX chain such as a USRP TwinRX
3) Distributed receiver (under what situation might you want to link them if at all?)

Does this change at all with the json-ld proposal in #108 and would there be any relation to linked recordings if one of them drops samples such as in #89?

bhilburn:
@jacobagilbert - Per our discussion yesterday, I know you've put a lot of thought into this particular PR. Do you mind outlining a list of the improvements you'd like this PR to cover and drive updating the code in this PR as needed? If so, I will assign to you 🙂

extensions/multirecordings.sigmf-ext.md
```diff
@@ -0,0 +1,103 @@
...
+The `recording` datatype is the primary mechanism by which SigMF Recordings are
+linked. This datatype is the name of a SigMF Recording, **without** extensions
+(i.e., given a recording comprising `N.sigmf-meta` and `N.sigmf-data`, the
+`recording` datatype value would simply be `N`).
+
+The metadata and dataset files of the recording MUST be in the local directory
+(i.e., local to the SigMF Recording in which they are referenced, and thus in
+the same SigMF Archive).
+
+### `core` Namespace Support
```
jacobagilbert, line 40: Is this necessary and worth it to introduce the concept of allowing extensions to modify `core`? We discussed this elsewhere and think we both are of the opinion that it is not. Separate `captures` segments can already have variable frequency metadata.

extensions/multirecordings.sigmf-ext.md
```diff
@@ -0,0 +1,103 @@
...
+
+#### Global Object
+
+| name          | required | type      | description                         |
+| ------------- | -------- | --------- | ----------------------------------- |
+| `master`      | false    | recording | The primary recording this recording is linked to.|
+
+The `multirecordings:master` field in the `global` object allows the user to
+indicate another SigMF recording that acts as the 'master' or 'parent' of this
+one, effectively providing a mechanism for bi-directional link between SigMF recordings.
```
jacobagilbert, line 67: Should we add language here to the effect:

> If this field is *not* included, it is assumed that this file is the master or sole recording

extensions/multirecordings.sigmf-ext.md
```diff
@@ -0,0 +1,103 @@
...
+
+#### Capture Segment Objects
+
+| name           | required | type      | description  |
+| -------------- | -------- | --------- | ------------ |
+| `streams`      | false    | object    | List of SigMF recordings that represent multiple streams.|
+
+The `multirecordings:streams` field in a capture segment object is a JSON array
```
jacobagilbert, line 75: How does this work if there are multiple segments at nonzero offsets? What does that imply for channels other than 'this' channel?

jacobagilbert (changes requested):
This is a really essential capability and definitely should be in v1.0.0. Also, there are a ton of formatting things here that could probably be separated into their own PR...but not a big deal.

jacobagilbert:
@bhilburn in terms of additions, i think adding an int-type `channel` field to the `global` object might be a good idea; its for specific use cases, but I believe very common ones (phase coherent recordings, linking multiple sensor nodes, etc.

extensions/multirecordings.sigmf-ext.md
bhilburn: Agreed, there should be no changes to `core` by extensions.

extensions/multirecordings.sigmf-ext.md
bhilburn: @jacobagilbert - I don't think I understand the gap you're pointing out. Can you explain it in a bit more detail for me?

bhilburn:
> 
> 
> @bhilburn in terms of additions, i think adding an int-type `channel` field to the `global` object might be a good idea; its for specific use cases, but I believe very common ones (phase coherent recordings, linking multiple sensor nodes, etc.

I like this idea, and agreed.

extensions/multirecordings.sigmf-ext.md
jacobagilbert: Sure - I think my question here has to do with this `streams` object being part of a `captures` segment, which starts at a user-defined sample (not necessarily 0) and extends until the next segment or EOF. That seems like something that might be a challenge. It also might be well thought out, so this is not to say 'this is a bad idea' -- rather 'is this what was intended'? Or perhaps this should be in `global`?

bhilburn: I removed this from my re-write of the extension, and then realized why it was there 😅

This is needed to point from the primary recording down to children. That said, I agree with your comment about it being in captures. The new PR moves this to global.

bhilburn:
This PR is replaced by #151

PR #101, opened 2018-10-11
Add DOI fields to `global`

bhilburn:
This PR adds fields for providing registered DOI strings for the metadata and dataset files. We've discussed this in the past, including briefly at GRCon, and we are already using DOIs for (our only) SigMF releases.

I chose to do this at the level of the individual files instead of at the level of Recordings (`N.sigmf-meta` & `N.sigmf-data` together) because we have already established support for, and confirmed the utility of, splitting the files and being able to treat them independently. For example, someone may create a recording, then process it and add annotations to the metadata file and share their work. Later, someone else might do further processing, modifying the metadata file, and in their publication need to cite the original dataset file.

This does raise the question of carrying NOTICEs, which we haven't discussed directly before. For example, if someone shares a SigMF recording under the CC-BY license, thus requiring carrying notice, in the example above where someone is providing a new metadata file for the same dataset file, how do they carry that notice? There isn't a good way to do that in the metadata, currently. The best solution might be specifically calling out a NOTICES file for recordings. 

Please share your thoughts & feedback on this PR and the issue of notice raised above!

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -31,29 +31,35 @@ This document is available under the [CC-BY-SA License](http://creativecommons.o
 <!-- markdown-toc start - Don't edit this section. Run M-x markdown-toc-generate-toc again -->
 **Table of Contents**
 
-- [Signal Metadata Format Specification v0.0.1](#signal-metadata-format-specification-v001)
-    - [Abstract](#abstract)
-    - [Status of this Document](#status-of-this-document)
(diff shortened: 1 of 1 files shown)


bhilburn:
Realized there was really no need for the new subsection in the spec describing the DOI. That's really more appropriate for the Wiki, which as we already discussed at GRCon, we need anyway.

Pushed a new commit that removes the extra language.

Issue #103, opened 2019-03-27
Extensions list should be an array


dboulware:
SigMF describes the extensions as a "JSON array of name/value pairs describing SigMF Extension namespaces, where the name is the namespace provided by an extension and the value is a string that specifies whether the extension is optional or the version of the extension required to properly parse & process the SigMF Recording." However, in the example that is shown, it is an object with key value pairs where the key is the namespace provided by an extension and the value is a string that specifies whether the extension is optional or the version of the extension required to properly parse & process the SigMF Recording.  When binding the JSON to objects, this requires providing a custom serializer and deserializer. If it were just changed to something like:
"core:extensions" : [
       { "name": "extension-01", "version": "optional" },
       { "name": "extension-02", "version": "v1.2.3" }
   ]
you would not have to create a custom serializer and deserializer.

bhilburn:
Hey @dboulware! Thanks so much for opening this.

A few things:

1. Great catch regarding the mismatch between describing it as an array, but defining it as an object! @n-west and I both agree with you that it should be an array. I'll include this in a PR I'll put together, along with some other changes listed below.

2. I don't think I follow the bit about a custom serializer / deserializer. If you paste the following:
```{"core:extensions" : {
      "extension-01": "optional",
      "extension-02": "v1.2.3"
    }}
```
Into [this online JSON parser](http://json.parser.online.fr/), it's valid JSON. We are going to move this over to an array, anyway, per the first item, but I did want to double-check this to make sure I full understood the issue in case it's still relevant.

There is a third piece to this, too. We discussed this at GRCon18 in the SigMF working group, but the list of extensions should really provide a URI to a machine-readable JSON schema that defines extensions. This would make it possible for SigMF tools to automatically pull and parse full extensions. I'll handle this in a separate issue.

djanderson:
@bhilburn looks like you had a copy/paste issue above.

Were you suggesting something like:

```json
[
    {"extension-01": "optional"},
    {"extension-02": "v1.2.3"}
]
```

Although really concise notation, I agree with what @dboulware says about that harder to parse for many JSON serializer/deserializer libraries. Instead of having a variable key, using known keys with known value types tends to be a lot easier.

And while we're discussing this part of the spec, I think it would even be worth it to break out "optional-ness" from the version. What if you want to specify the version of something that's optional?

Recommended:

```json
{
    "extensions": [
        {
            "name": "extension-1",
            "version": "v0.3",
            "optional": true
        },
        {
[...]

PR #105, opened 2019-04-01
Modifications to `extensions` per #103.

bhilburn:
This branch makes the changes described in #103, including:
1. Move the extensions list to an actual JSON array
2. Break version out from optional.

I don't really care for the naming of the fields in the `extensions objects` at the moment. I first made them without the `ext_` prefix (e.g., `core:name`), but then decided against it as it creates a name conflict with `core:version` in the parent global object. I'd love to hear thoughts & feedback on this topic.

Note this PR also contains a few other issues to fix some MD linter warnings, which is a bit sloppy to wrap up in this branch - my apologies.

### DIFF
--- a/extensions/adsb.sigmf-ext.md
+++ b/extensions/adsb.sigmf-ext.md
@@ -1,15 +1,20 @@
 # ADS-B Extension v1.0.0
-The Automatic Dependent Surveillance-Broadcast (`adsb`) namespace extension defines dynamic properties of ADS-B signals extending `annotations`.
 
-## Global
-The following names are specified in the `adsb` namespace and should be used in the `global` object:
+The Automatic Dependent Surveillance-Broadcast (`adsb`) namespace extension
+defines dynamic properties of ADS-B signals extending `annotations`.
 
-|name|required|type|unit|description|
-|----|--------------|-------|-------|-----------|
-|`version`|true|string|N/A|The version of the `adsb` extension used to create the metadata file.|
+## 1 Global
 
-## Annotations
-The following names are specified in the `adsb` namespace and should be used in the `annotations` object:
+`adsb` does not extend [Global](https://github.com/gnuradio/SigMF/blob/master/sigmf-spec.md#global-object).
+
+## 2 Captures
+
+`signal` does not extend [Captures](https://github.com/gnuradio/SigMF/blob/master/sigmf-spec.md#captures-array).
+
+## 3 Annotations
+
+The following names are specified in the `adsb` namespace and should be used in
+the `annotations` object:
 
 |name|required|type|unit|description|
 |----|--------------|-------|-------|-----------|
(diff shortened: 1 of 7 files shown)


djanderson:
I actually don't think we need to append `core:` _or_ `ext_` to the keys in the extension object. The `core:extension` key puts the entire object in the `core` namespace and the key names are only allowed within that object, not at the top level, so they'll never conflict with existing fields. 

I think it's common practice to reuse key names in distinct objects.

n-west:
That decision should probably be bigger than just the extension object fields. Right now I don't know of any other field that has an `object` type for value. I think I would lean away from names that cannot be directly used as variable names in common languages (python, c, c++) but we've already crossed that threshold. I think there should be some higher level guidance that values of object type either always use the parent namespace or that object values don't use a namespace.

I need to look at my C++ serializing tools to figure out what would be most sane to support before offering further suggestions, but even that would just be based on my preference because of tools I already have that I'd rather not be broken by a decision like this :-).

djanderson:
Here's an example to make my point:

```python
import json


class NamespaceExtension(object):
    """Type used deserialize core:extensions objects."""
    def __init__(self, name, version, optional):
        self.name = name
        self.version = version
        self.optional = optional


# Some example metadata
json_md = """{
    "global": {
        "core:extensions": [
            {
                "name": "extension-01",
                "version": "v0.0.5",
                "optional": true
            },
            {
                "name": "extension-02",
                "version": "v1.2.3",
                "optional": false
            }
        ]
    }
}"""


if __name__ == '__main__':
    md = json.loads(json_md)
    for ext in md["global"]["core:extensions"]:
        try:
            deserialized_ext = NamespaceExtension(**ext)
        except Exception as err:
            print("Invalid core:extensions object - {}".format(err))

        fmt = "Extension {0.name} ({0.version}): optional: {0.optional}"
        print(fmt.format(deserialized_ext))
```
Outputting:
```bash
Extension extension-01 (v0.0.5): optional: True
Extension extension-02 (v1.2.3): optional: False
```
Note that if we prepend `core:` to the names, they would not map cleanly to variable names that we can use in most common languages (as @n-west mentioned). There is no chance of confusing the keys inside the `core:extensions` object with keys in other object named the same thing, so adding `ext_` doesn't add any valuable information. Try rewriting the code above to use `ext_` and it will be either harder to deserialize or not read as cleanly.

bhilburn:
Okay, I think the resolution, here, is to update this PR to not prepend `core` to every subfield, and that it should look like the code in @djanderson's snippet, above. Specifically:

```json
{
    "global": {
        "core:extensions": [
            {
                "name": "extension-01",
                "version": "v0.0.5",
                "optional": true
            },
            {
                "name": "extension-02",
                "version": "v1.2.3",
                "optional": false
            }
        ]
    }
}
```

Will tag this for me

jacobagilbert (approved):
This is a solid improvement.

gmabey:
@bhilburn Is this PR truly ready to merge?

bhilburn:
@gmabey - No, I have not made the changes referenced in my previous comment, yet.

jacobagilbert:
@bhilburn this has been rebased and updated per discussion above. The last thing we need to do before merge is to remove the individual `global` scope `version` fields in the canonical extensions. I am in favor of completely removing them in v1.x (technically not backward compatible with v0.x but 'safe' still as even if they were used they will just be ignored). That OK w/ you?

Are `adsb`, `rfml`, and `wifi` considered canonical extensions? Or just extensions included here for convenience? I'll add them in if they are.

Whats up with `volatile`? Its just an empty file and I don't see a need to keep this around in its current state.... Seems like `multirecordings` is actually going to cover what this was intended for?

bhilburn:
@jacobagilbert - Oops, thought I had already responded to this.

* volatile is a ghostfile, will be removed in #151 merge.
* All other extensions /are/ canonical and should be updated accordingly.

If you don't mind updating them / adding them to this PR, we should be good, here

jacobagilbert:
@bhilburn this dropped off my queue... but should be good to go now. I also took this opportunity to add section numbering to each canonical extension.

PR #107, opened 2019-05-17
Fix citation and license

dbouquin:
Also changed the tar file requirement. Will create issues for other bugs.


bhilburn:
Thanks for this PR, @dbouquin!

I'm torn about what to do regarding the `png`s. Including them in the repository directly without using `git-lfs` creates binary bloat within the repo - it won't be bad if we never touch those files again, but if we plan to update them, could become problematic.

The primary issue with `git-lfs` is that if you, as a user, are unfamiliar with how to use it, then it can be a barrier to properly cloning the entire repository.

Another option is to upload the images to the repo through issues / PRs, and then use their permanent URLs from documentation / the specification. This works well in other projects, but it does mean that the images are not included in the repository, proper (i.e., there is no way to grab them with a `git clone`).

Currently, I'm leaning towards [git-lfs](https://git-lfs.github.com/), but am happy to hear other opinions / ideas?

bhilburn:
Assigning this to v0.1.0

Teque5:
For the images you should use `svg` where possible so your images are <5KB for simple cases like this. Otherwise I think the only okay use of images is for a logo or something else long-standing.

We tried `git-lfs` in my shop but eventually decided [dvc](https://dvc.org/) was much easier to use for tracking files within git repos. In either case you need hosting for the data so YMMV.

bhilburn:
I'd like to get this PR updated for v1.0.0 and get it merged. The things I want to do:

* change version and credit in citation file
* probably change license type from a share-alike to just a CC-BY for the spec
* specify a license for the code to Apache 2.0
* clearly call out inbound=outbound for contributing (no CLA required)

PR #111, opened 2019-06-29
Require metadata to precede data member

paultag:
Requiring the metadata to come before the data member allows you to process the archives data without actually unpacking it to the filesystem or copy all data into memory.

We take a similar approach (but with the older `ar` format not `tar`) in Debian, where the order must be the version member, the `control.tar.X` member, and finally the `data.tar.X` member.

This has no effect to any end-users who are manually unpacking the archives, nor any code which currently accepts any order of members, but it will mean new parsers that rely on this ordering will horck on existing archives. As such this is perhaps (if you squint just right) a breaking change in the spec, and I'm not sure how that is managed within this community. Feedback is very welcome.

Thanks for maintaining this spec!

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -135,10 +135,11 @@ into a file archive. A `SigMF Archive` may contain multiple `SigMF Recordings`.
 3. The archive MUST contain the following files: for each contained recording
    with some name given here meta-syntactically as `N`, files named `N` (a
    directory), `N/N.sigmf-meta`, and `N/N.sigmf-data`.
-4. The archive MUST NOT contain any other files unless their pathnames begin
+4. The `N/N.sigmf-meta` MUST come before the `N/N.sigmf-data` member.
+5. The archive MUST NOT contain any other files unless their pathnames begin
    with `N/N`, for some `N` which has `.sigmf-meta` and `.sigmf-data` files as
    described above.
-5. It is RECOMMENDED that if recordings in an archive represent a continuous
+6. It is RECOMMENDED that if recordings in an archive represent a continuous
    dataset that has been split into separate recordings, that their filenames
    reflect the order of the series by appending a hyphenated zero-based index
    (e.g., `N-0`, `N-1`, `N-2`, etc.,).


smunaut:
And requiring that prevents writing the file without having to buffer the data somewhere until you can write it ...  since metadata are smaller it's much better to keep those and directly write the data into the archive and then when done, flush the metadata at the end.

sigmf archive are tar, not tar.gz or whatever compressions format and as such, for reading you can freely seek in them.

paultag:
I suppose this is all tradeoffs. Being able to read without unpacking is very attractive to consumers at the expense of producers, since you can just read the stream of data as needed and never sync it to disk.

Not doing this makes things easier for recorders, to your point -- especially with annotations. You would know most of it up front, but not all of it, to your point.

I don't know either option is good. Relying on the archive being on disk in order to parse it isn't ideal.

My issue here is basically consumers of the data will, in a mirror to your consumer story, have to either store all the very large data in memory to get the very small metadata, or download data to the client disk and either unpack or seek + read repeatedly.

However, to your point, pretending to a file isn't practical to do in a repeatable way.

I'll think about this more.

smunaut:
HTTP supports segment fetch in any order, you can seek in most modern network protocols.

paultag:
I think /most/ is a stretch. But http yes.

bhilburn:
@paultag - Sorry it's taken me a few days to pull this up! 

I think I'm actually having trouble understanding the changeset in the context of the discussion here, I think. The only real change in the diff is:

```ascii
The `N/N.sigmf-meta` MUST come before the `N/N.sigmf-data` member.
```

But I'm not sure how to interpret that from the context of archives in the discussion. Correct me if I'm wrong, but my impression is that object order in archives is practically random, and even if it wasn't, I'm not sure I understand how "ordering" affects processing?

I think there must be something fundamental here I'm just completely ignorant of or misunderstanding, hah. Can someone point me in the right direction? =)

paultag:
@bhilburn absolutely! 

The archive from a bytes-perspective is a header which includes metadata, and the length of the file, then the file's bytes. After the file is another header, and so on.

If you're reading a file entry by entry, you can read the header bytes, process metadata, and then process the file bytes. This works great when you want to extract all files, or maybe extract a single or specific file. When that file is on a filesystem, you can `seek` to the next offset using the file length if you don't want to process that file yet, and move to the next one, and then seek back to another offset to read another member.

The issue I'm trying to solve here is when you consider a stream without `seek` (as many network protocols do), or streams that incur a penalty on seeks / random access.

This ordering will allow you to read each member in order without using `seek` on a read only (maybe even broadcast!) stream, and store the metadata in memory while processing each file.

This means that when you get to IQ data, you will have the metadata instructing your code on how to process that information (meaning you could write demodulated output rather than raw IQ), without having to write the IQ data to disk while waiting for the metadata instructing you on how to interpret the bytes.

As @smunaut argues: This does make writing archives harder if you're directly writing archive format, since you will sometimes have to re-write or update metadata during capture (annotations, frequency change), which means rather than storing metadata in memory and flushing to disk at the end of the capture, you'd have to do write the IQ to disk then form the archive (which may even take more disk space since you'll store IQ twice) after the metadata is in memory.

My thoughts on their feedback is -- while I didn't think anyone was relying on it, I do understand their point, as well as the fact this may be a spec incompatibility. My use-case may be exotic and I can handle it through a different mechanism (e.g. just making a static sized binary header with the format information and avoiding SigMF when transmitting, and use it as a compatibility mechanism on export / translation later), by treating SigMF more like a compatibility format rather than a native format. 


Is that a bit more clear or am I still having a tough time explaining the odd train of constraints I've found myself under

paultag:
(This is also to say, I don't mind if this is a no-go, I just wanted to have the discussion and ensure it got a review)

bhilburn:
@paultag - Ah, okay, I think I follow a bit better, now. Thanks so much for the thorough write-up! It's much appreciated.

Yeah, so the point @smunaut raises goes back to the original design requirements for SigMF. It's very specifically designed to be a storage format, not an over-the-wire / network / packet format. So design decisions are optimized assuming an application streaming samples and/or annotations to disk, or a reader streaming them off-disk (or `mmap`ing a file on-disk and reading). 

I have a better understanding of what you are trying to achieve now, but, requiring it would make high-rate annotation recordings more difficult & intensive since you would have to hold all of it in memory. I don't see a good way to both enable the use-case you are describing in a meaningful way (i.e., making it a canonically required format) without directly complicating the primary goal (high-rate storage format).

Does my reasoning above make sense in the context of your explanation?

paultag:
Sure does! Thanks for the discussion! ➕

PR #106, opened 2019-05-08
new i/o methods for annotations & samples; +python3

Teque5:
The following squashed commit was approved for public release by **The
Aerospace Corporation** on 2019-05-07. It is covered software release
request #SW19-0024. Commits made by the *Digital Communications
Implementation Department*.

New features
- method to read & write SigMF metadata files
- method to read available captures
- method to read samples from captures
- method to get annotations
- method to count samples
- python3 module compatibility
- examples added to README

Fixes
- python3 relative path fixes
- bug in test_validation
- bug in conftest that used old SigMF datatype (f32)
- additional test in test_sigmffile
- made compatible with docker containers
- can now write multiple annotations for same sample start index

Pylint code cleanup
- removed unused imports
- fixed indentation
- fixed snake_case and too short variables
- fixed missing error definition

Other changes
- Remaining core namespace keys were added as constants to SigMFFile class.
- Sample count is automatically determined and set as a class data member
  when a data file is present.

PR #115, opened 2020-06-15
Feature/local gui sigmf archive file editing

nathan-turner111:
This branch is based on a PR by Teque5 called 'new i/o methods for annotations & samples; +python3'. I have not modified any code by the user, I just used some of the new functionality they added.

python sigmf_gui.py

This feature adds a GUI for users to edit, share, and load others' signal collections in a common formal. The GUI allows users to open a SigMF archive file, or create their own by adding a binary data file and saving it with annotations and signal information. Users can also open other users' archive files, which will populate the GUI with information on the collected signal. The user can edit the data and resave the archive. Multiple captures within a file can be added and modified, so users can describe each capture in a data file. There are required fields for storing a valid archive file in red, and many other useful fields that others may want to know about when looking at a signal collection. Hopefully this helps users share data and easily record their collections.

### DIFF
--- a/scripts/sigmf_gui.py
+++ b/scripts/sigmf_gui.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 # Copyright 2020 The Johns Hopkins University Applied Physics Laboratory LLC.  All Rights Reserved.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -140,7 +141,7 @@ class WindowInput(WindowElementGroup):
         el_help = {WindowInput.DATA_BYTE_ORDER: 'Data Type Help'}
         el_multiline = [DESCRIPTION]
         el_selector = {WindowInput.DATA_SAMPLE_SIZE: [8, 16, 32],
-                       WindowInput.DATA_BYTE_ORDER: ['big endian', 'little endian']}
+                       WindowInput.DATA_BYTE_ORDER: ['little endian', 'big endian']}
         el_checkbox = [WindowInput.DATA_TYPE_COMPLEX, WindowInput.DATA_TYPE_UNSIGNED, WindowInput.DATA_TYPE_FIXEDPOINT]
         el_size = {WindowInput.DATA_TYPE_COMPLEX: (3, 1), WindowInput.DATA_TYPE_UNSIGNED: (3, 1),
                    WindowInput.DATA_TYPE_FIXEDPOINT: (3, 1), WindowInput.DATA_SAMPLE_SIZE: (4, 1),
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ setup(
     ],
     packages=['sigmf'],
     package_data = {'sigmf': ['*.json']},
-    install_requires=['six', 'numpy'],
+    install_requires=['six', 'numpy', 'pysimplegui==4.0.0'],
     setup_requires=['pytest-runner'],
     tests_require=['pytest>3'],
     zip_safe=False
--- a/sigmf_gui.py
+++ b/sigmf_gui.py
@@ -1,0 +1,606 @@
+# Copyright 2020 The Johns Hopkins University Applied Physics Laboratory LLC.  All Rights Reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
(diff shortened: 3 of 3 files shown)


bhilburn:
Okay, it's late, so grains of 🧂 and all, but I'm having trouble running this from the combined #119. Get this when trying to run:

```
$ python sigmf_gui.py
Exception ignored in: <_io.FileIO name='/home/bhilburn/code/sigmf.git/sigmf/schema.json' mode='rb' closefd=True>
  File "/home/bhilburn/code/sigmf.git/sigmf/schema.py", line 32, in get_schema
    return json.load(open(schema_file))
ResourceWarning: unclosed file <_io.TextIOWrapper name='/home/bhilburn/code/sigmf.git/sigmf/schema.json' mode='r' encoding='UTF-8'>
  File "sigmf_gui.py", line 606, in <module>
    run_gui()
  File "sigmf_gui.py", line 473, in run_gui
    validate_button.Update(text='Update')
  File "/home/bhilburn/.local/lib/python3.8/site-packages/PySimpleGUI/PySimpleGUI.py", line 3380, in Update
    if not self._widget_was_created():      # if widget hasn't been created yet, then don't allow
  File "/home/bhilburn/.local/lib/python3.8/site-packages/PySimpleGUI/PySimpleGUI.py", line 1055, in _widget_was_created
    warnings.warn('You cannot Update element with key = {} until the window has been Read or Finalized'.format(self.Key), UserWarning)
UserWarning: You cannot Update element with key = Update until the window has been Read or Finalized
Error in sys.excepthook:
  File "/usr/lib/python3/dist-packages/apport_python_hook.py", line 72, in apport_excepthook
    from apport.fileutils import likely_packaged, get_recent_crashes
  File "/usr/lib/python3/dist-packages/apport/__init__.py", line 5, in <module>
    from apport.report import Report
  File "/usr/lib/python3/dist-packages/apport/report.py", line 13, in <module>
    import fnmatch, glob, traceback, errno, sys, atexit, locale, imp
  File "/usr/lib/python3.8/imp.py", line 31, in <module>
    warnings.warn("the imp module is deprecated in favour of importlib; "
DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses

Original exception was:
  File "sigmf_gui.py", line 606, in <module>
    run_gui()
  File "sigmf_gui.py", line 473, in run_gui
    validate_button.Update(text='Update')
  File "/home/bhilburn/.local/lib/python3.8/site-packages/PySimpleGUI/PySimpleGUI.py", line 3380, in Update
    if not self._widget_was_created():      # if widget hasn't been created yet, then don't allow
  File "/home/bhilburn/.local/lib/python3.8/site-packages/PySimpleGUI/PySimpleGUI.py", line 1055, in _widget_was_created
    warnings.warn('You cannot Update element with key = {} until the window has been Read or Finalized'.format(self.Key), UserWarning)
UserWarning: You cannot Update element with key = Update until the window has been Read or Finalized
```

Rolling with Python 3.8.2 on Ubuntu 20.04

@nathan-turner111 - Have you seen anything like this

nathan-turner111:
@bhilburn are you still seeing this issue, or were you able to fix it in #119? I have not seen this error before and was able to checkout #119 and use the GUI without issue. I'll spend some more time trying to reproduce the problem if you are still running into the error.

bhilburn:
With #106 and #116 merged, this PR now has conflicts - it's the issue with shared history mentioned [here](https://github.com/gnuradio/SigMF/pull/119#issuecomment-683227741). @nathan-turner111 - Can be easily fixed by re-basing this branch on top of master, or cherry-picking the commit from your branch over to master and updating this PR. 

Some good comments from @Teque5 over in #119, copy / pasted here:

- Perhaps this should be in a scripts/ folder that has a link to it through the setup.py as either a script or entry_point. It seems odd that it's in the root.
- I think little endian should be default. it's currently set to big-endian.
- I can't figure out how to load a sigmf file. It seems I can load an archive, but not a pair of .sigmf-data and .sigmf-meta files.
- I ran it with python3 sigmf_gui.py, but you should really just add #!/usr/bin/env python3 and set it as executable.

One of @Teque5's comments for discussion:

- I'm not sure I see the point of the GUI. If I started with only a data file and wanted to create an archive this would be useful but I usually start with the gr-sigmf output and already have metadata.

You're right that if you already have a full recording, the GUI isn't so useful. There are a number of usecases / workflows where that's not true, though, and I think this is really helpful. I see this as a good starting point for a tool that can expand to cover other workflows and be really useful for managing recordings. Thoughts, @nathan-turner111?

nathan-turner111:
Thanks @bhilburn, I'll rebase my branch now. I'll also address the comments from @Teque5. I agree the GUI wasn't made for users that already have a tagged data collection. The largest benefit I see is for users that have an untagged binary collection that makes it difficult to share with other projects. Instead of the other projects having to collect their own data with the tags they require, the GUI makes it easier to save the data in a common format and encourages users to fill out as many tags as possible. As for the comment about loading SigMF files other than the archive file, I think this may need to be a follow on feature. I no longer have the ability to work on this project. The GUI currently supports loading binary data and archive files, but there are many other collection formats that should be supported in the future.

PR #116, opened 2020-08-20
ordered metadata; improved setup; better use cases in readme

Teque5:
This squashed commit is approved for public release by **The Aerospace Corporation** on 2020-08-19. Commits made by the _Communication Software Implementation Department_.

This commit is an addendum to the prior [PR#106](https://github.com/gnuradio/SigMF/pull/106) approved 2019-05-07.

### New Features
* Ordered sigmf-meta so that the keys are in a more useful order upon manual inspection.
* Use cases in `README.md`.

### Fixes
* Improved version handling.

### Notes
When the metadata is written to a file it now looks like this so humans can read it with `global` followed by `captures` followed by `annotations`. Metadata below is created by the new example in the `README.md`.
```json
{
    "global": {
        "core:author": "jane.doe@domain.org",
        "core:datatype": "cf32_le",
        "core:description": "All zero example file.",
        "core:sample_rate": 48000,
        "core:sha512": "18790c279e0ca614c2b57a215fecc23a6c3d2d308ce77f314378cb2d1b0f413acd3a9cd353aa6da86ec9f51916925c7210f7dfabc0ef726779f8d44f227f03b1",
        "core:version": "0.0.2"
    },
    "captures": [
        {
            "core:datetime": "2020-08-19T23:58:58.610246Z",
            "core:frequency": 915000000,
            "core:sample_start": 0
        }
    ],
    "annotations": [
        {
            "core:comment": "example annotation",
            "core:freq_lower_edge": 914995000.0,
            "core:freq_upper_edge": 915005000.0,
            "core:sample_count": 200,
            "core:sample_start": 100
        }
    ]
}
```

PR #119, opened 2020-08-29
Staging Updates to Python Code from Aerospace Corp.

bhilburn:
This PR stages #106 and #116 onto `SigMF/master`.

I am staging these here because they are both rather large changesets, and the latter builds on the former, so I don't think it makes sense to evaluate #106 on its own at this point - especially since it was filed over a year ago.

@Teque5 - I confirmed that this staging branch correctly has your two changesets by `diff`ing it against `aerospace/public/release`. The only difference is that you originally branched before I merged `CODE_OF_CONDUCT.md` into `master` (see 336fcec398b), and you haven't pulled since then.

Update: have now also cherry-picked #115 to this branch.

Closes #106, #116, #115.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,15 @@
+# temp files
 *.swp
 *.pyc
 .cache
+
+# setuptools related
+build/*
+.eggs/*
+SigMF.egg-info/*
+
+# pytest & coverage related
+.coverage
+pytest.xml
+coverage.xml
+htmlcov/*
--- a/README.md
+++ b/README.md
@@ -39,28 +39,22 @@ maintained for posterity.
 Anyone is welcome to get involved - indeed, the more people involved in the
 discussions, the more useful the standard is likely to be.
 
-## Installation
-After cloning, simply run the setup script for a static installation.
+## Getting Started
 
-```
-python setup.py
-```
-
-Alternatively, install the module in developer mode if you plan to experiment
-with your own changes.
-
-```
-python setup.py develop
+This module can be installed the typical way:
+```bash
+pip install .
 ```
 
-## Usage example
-#### Load a SigMF dataset; read its annotation, metadata, and samples
+## Use Cases
+### Load a SigMF dataset; read its annotation, metadata, and samples
+
 ```python
 from sigmf import SigMFFile, sigmffile
 
 # Load a dataset
-sigmf_filename = 'datasets/my_dataset.sigmf-meta' # extension is optional
-signal = sigmffile.fromfile(sigmf_filename)
+filename = 'example.sigmf-meta' # extension is optional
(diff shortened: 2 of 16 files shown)


bhilburn:
Okay, I just cherry-picked @nathan-turner111's commit from #115 over to this branch. PR #115 and #116 shared a common history in changeset but not commits, so merging them both would have turned into a merge conflict nightmare.

@nathan-turner111 @Teque5  - This PR should now include all 3 PRs that you've filed, and merging this branch will close all three. Focusing review / edits here to organize the effort.

bhilburn:
By the way, now that I've slept, #115 would have been fine, hah. Still planning to test it on top of #106 and #116 since I'm guessing that's how its being used at Aerospace. If that causes issues, we can split it back out, and then rebase it once #106 and #116 are merged 🙂

bhilburn:
@Teque5, @nathan-turner111 - If you're able, it would be *super* helpful if you could pull this staged branch and test it to make sure it works as you expect as compared to your fork. Assuming it does, we should be good. If not, we just need to resolve the discrepancies.

Tagging @dkozel, as well, as I know he has an interest in getting this done too.

bhilburn:
Just a ping on this, @Teque5! Any chance your team can give this branch a go? I believe it should be exactly what you are using in the branch on your fork.

nathan-turner111:
@bhilburn - I just pulled the branch and ran the GUI. It works as expected for me. However after some testing I realized I was not able to get the GUI running with python 3.6, so at least python 3.7 will be required to use it for now. I also noticed that the newest version of numpy requires python 3.5, but the project is classified under many versions of python below 3.5. I can't speak for @Teque5 commits.

Teque5:
I will take a look at this in a few hours.

Teque5:
So all the code from our PR looks fine and works with what we have. The GUI works just fine with python 3.6.9 on my Ubuntu 18.04 machine in a virtualenv. But I do have some notes:

- Perhaps this should be in a `scripts/` folder that has a link to it through the `setup.py` as either a `script` or `entry_point`. It seems odd that it's in the root.
- I think little endian should be default. it's currently set to big-endian.
- I can't figure out how to load a sigmf file. It seems I can load an archive, but not a pair of `.sigmf-data` and `.sigmf-meta` files.
- I'm not sure I see the point of the GUI. If I started with only a data file and wanted to create an archive this would be useful but I usually start with the `gr-sigmf` output and already have metadata.
- I ran it with `python3 sigmf_gui.py`, but you should really just add `#!/usr/bin/env python3` and set it as executable.

bhilburn:
#106 and #116 are now merged. Going to move comments from here regarding the GUI over to #115

Issue #84, opened 2017-11-28
'altitude' field, and being able to put coordinates in global


777arc:
I reviewed the current version of SigMF to see if it would work for what I'm doing, and the only thing that would be nice to add is an optional altitude field after the latitude and longitude.  I mean it's optional so people without altitude can just leave it empty.

Another aspect that seems a little weird is that I can't just put the coordinates in global, which applies to the entire dataset.  For what I'm doing there's a set of coordinates that goes along with every file.  With the current SigMF I suppose I can just add an annotation to the first sample in each file, with a sample_count of either 1 or the number of samples in the dataset.  I guess that works, it just feels a little weird.  I think a lot of applications involve saving coordinates that don't change over a file and may never change =)

Also I think it was already mentioned but we need to decide on a standard coordinate system/units.  Doesn't matter, just needs to be standardized to save headache.

bhilburn:
@777arc - Check out #73! We came up with the exact same suggestion :) Be sure to check out the discussion, there, and comment.

Now, the one thing that is still undecided is how, exactly, we describe geolocation. Ideally we identify a standard that includes not just lat/long but also altitude, somehow. I'm honestly not sure if that exists, though.

Do you have any knowledge of how to do this properly? Pinging @dharasty in case he has thoughts.

777arc:
I would say go with what KML uses, which is decimal degrees for lat/lon, and meters for altitude.  You also need an altitude-mode that chooses between:

1- distance from the surface of the Earth (relativeToGround)
2- distance above sea level (absolute)
3- distance from the bottom of major bodies of water (relativeToSeaFloor)

So four fields total

bhilburn:
Will address in v0.0.2

bhilburn:
@777arc  - any chance you can recommend another published standard / RFC / etc., to reference for this?

jacobagilbert:
@bhilburn OGC KML (https://www.ogc.org/standards/kml/) is the name for what @777arc recommended and its pretty common and sane. I think this would be a good choice for the fix for this and #73 .

Issue #73, opened 2017-09-20
`lat` and `long` in Annotations


bhilburn:
Right now, `latitude` and `longitude` are in the `core` namespace for Annotations segments. This was put in *before* we had figured out the continuously variable method for annotating geolocation (per #55).

I think this opens the question, do we still want `lat` and `long` in the `core` namespace as a field in one of the segments? And, if so, does it belong in Annotations?

We *know* that many people will be using the new multiple-recordings approach to record geolocation for a recording. The downside, if that is the only way to define geolocation, is that we create a lot of overhead if you have a stationary receiver that you want to provide the coordinates for. The upside, though, is that there is no risk of confusion regarding what to do if both a Recording of geolocation and `lat` & `long` tags exist.

Here is my suggestion: In the `global` namespace, we add a `core:geolocation` field. This field can either be a formatted lat/long coordinate, in which case it's static for the recording, or it uses whatever method for linking to another recording that we land on in #55, in which case it's obvious that it's volatile.

Thoughts?

bhilburn:
It's worth noting that if we go with my suggested route for geolocation, we probably will want to do the same thing for antenna azimuth.

kpreid:
Approaching the problem from a different angle: have a way to write an annotation which explicitly is equivalent to a low-sample-rate secondary recording (specifically, two equal values at the endpoints of the annotation's time interval).

Thus, for each property of a recording that might want to vary over time, it can be represented using either a "link to another recording" or an "inlined constant recording".

mbr0wn:
What @kpreid says is good for genericity. For the very-low-rate recording (aka just a single value), I would strongly advise that we can simply write something human-readable into the metadata file. Many people won't be both moving their receivers *and* tracking GPS locations while they do so. Same goes for antenna-steering (simply because it's a non-trivial effort to accurately know azimuth etc. of the antenna at all times). For those cases, which will be the majority of cases, we don't want to add overhead, and simply allow writing some text into the metadata file.

If we use lat/long, it's obvious what to put into the fields. There is no single way of storing geolocation, though. There's probably already multiple ways of storing GPS location data, and hell, there's other things such as [what3words](https://what3words.com) out there.
What I'm suggesting is that if we store geolocation data, we don't call the field 'geolocation', but give it a name that reflects the format in which the data is stored.

bhilburn:
[...]

PR #120, opened 2020-10-02
Refactor Modulation Extension Namespace

Atheatos:
Refactor the modulation extension based on discussions from https://github.com/gnuradio/SigMF/pull/87:
 - rename the namespace to `signal` from `modulation`
 - add a free-form `label` field for a short-hand description of the signal under `signal:emission`
 - move the signal descriptor fields to an object `Detail`, nested under `signal:emission`
 - add `channel` to `Detail` so that the communication system's bandwidth channel can be notated
 - rename `system` to `standard` under `Detail`
 - add an object `Emitter`, nested under `signal:emission`, to describe the hardware driving the communication system
 - update examples

I also added a `class_variant` field to `Detail` to cover less straightforward modulations such as pi/4-DQPSK and GMSK. This wasn't discussed in the other PR or previous issues, but something like it is necessary for the signals that I have. I'm not sure it's the best approach since its contents are open to interpretation, however. For example, should its value be `G` or `gaussian` for GMSK? How about `pi/4` vs `π/4`? Maybe a separate variant field for each class, perhaps in a completely different extension?

Here's how it looks for pi/4-DQPSK in the example I added in the spec:
```json
"detail": {
    "type": "digital",
    "class": "psk",
    "order": 4,
    "symbol_variant": "differential",
    "class_variant": "pi/4"
}
```


Teque5:
Nice Job. This looks pretty well thought-out.

extensions/signal.sigmf-ext.md
```diff
@@ -0,0 +1,285 @@
...
+
+The field of communications is vast, and there may be communications systems that cannot be described using the names and fields described in this extension. If you need additional or different fields to describe a system, create a new extension that adds the necessary fields to the `signal` namespace and/or submit the new fields to be upstreamed into this specification.
+
+#### The `emission` Object
+
+`emission` objects contain name/value pairs that describe the modulations of communications systems. These objects MAY contain the following pairs:
+
+|name|required|type|description|
+|----|--------------|-------|----------|
+|`label`|false|string|e.g., wifi|
```
bhilburn, line 55: I know this field seems relatively straight-forward, but it would be good to specifically call out it's meant to be user-defined shortform. Maybe something like:

```txt
A high-level shortform description useful to human readers (e.g., "wifi", "LTE")
```

extensions/signal.sigmf-ext.md
```diff
@@ -0,0 +1,285 @@
...
+|`emitter`|false|[Emitter](signal.sigmf-ext.md#the-emitter-object)|Emitter details (manufacturer, geo coordinates, etc.)|
+
+##### The Emitter Object
+
+|name|required|type|description|
+|----|--------|----|-----------|
+|`seid`|false|uint|Unique ID of the emitter|
+|`manufacturer`|false|string|Manufacterer of the hardware used to emit the signal|
+|`latitude`|false|float|Latitude|
+|`longitude`|false|float|longitude|
```
bhilburn, line 66: We should probably follow the NTIA's example, here, and specify the units as "decimal degrees" for both long & lat. Near the top of @MichaelCotton's comment, [here](https://github.com/gnuradio/SigMF/pull/87#issuecomment-508749532)

bhilburn (commented):
Excellent job, @Atheatos, taking all of the discussion and ideas over nearly 3 years of PR hacking and making a useful PR out of it! Thanks so much!

I think with a couple of minor fixes (see comments), this is ready for merge 👍

Atheatos:
Happy to hear that it looks good; thanks for the reviews.

I don't have anything to add to your description of the `label` field, so I just added it verbatim. Since adding a "units" column wouldn't really make sense for that table, I just added it to the description.

bhilburn:
Would like to get review from @jacobagilbert / @n-west, then want to merge ASAP.

extensions/signal.sigmf-ext.md
```diff
@@ -0,0 +1,285 @@
...
+
+The field of communications is vast, and there may be communications systems that cannot be described using the names and fields described in this extension. If you need additional or different fields to describe a system, create a new extension that adds the necessary fields to the `signal` namespace and/or submit the new fields to be upstreamed into this specification.
+
+#### The `emission` Object
+
+`emission` objects contain name/value pairs that describe the modulations of communications systems. These objects MAY contain the following pairs:
+
+|name|required|type|description|
+|----|--------------|-------|----------|
+|`label`|false|string|A high-level shortform description useful to human readers (e.g., "wifi", "LTE")|
```
jacobagilbert, line 55: I believe the `label` field is extremely common and should actually be moved to core (we could also call this `description` to be consistent with `global`), but this is really something I think should be up a level and more generic.

extensions/signal.sigmf-ext.md
```diff
@@ -0,0 +1,285 @@
...
+|`detail`|false|[Detail](signal.sigmf-ext.md#the-detail-object)|Emission details (standard, modulation, etc.)|
```
jacobagilbert, line 56: I think `signal:emitter` should be directly within the annotations object.

extensions/signal.sigmf-ext.md
```diff
@@ -0,0 +1,285 @@
...
+|`emitter`|false|[Emitter](signal.sigmf-ext.md#the-emitter-object)|Emitter details (manufacturer, geo coordinates, etc.)|
```
jacobagilbert, line 57: Same as above, `signal:emission_detail` or `signal:emission` should be directly within the annotations object. This also removes one of the hierarchical levels making this a bit simpler.

extensions/signal.sigmf-ext.md
```diff
@@ -0,0 +1,285 @@
...
+
+##### The Emitter Object
+
+|name|required|type|description|
+|----|--------|----|-----------|
+|`seid`|false|uint|Unique ID of the emitter|
+|`manufacturer`|false|string|Manufacterer of the hardware used to emit the signal|
+|`latitude`|false|float|Latitude (decimal degrees)|
+|`longitude`|false|float|Longitude (decimal degrees)|
```
jacobagilbert, line 66: Not to delay merging, but I think we should come to a consensus on how we handle geospatial metadata before we merge this. If we are going to use GeoJSON we should just put a `location` field here and allow the same format.

If we end up sticking with discrete coordinate fields, i suggest we add `altitude` and define the format.

See #73 and #84

extensions/signal.sigmf-ext.md
```diff
@@ -0,0 +1,285 @@
...
+The `class` field can have the following values:
+
+|value|description|
+|----|-------|
+|`am`|(analog) amplitude modulation|
+|`fm`|(analog) frequency modulation|
+|`pm`|(analog) phase modulation|
+|`ssb`|(analog) single side-band|
+|`dsb`|(analog) dual side-band|
+|`vsb`|(analog) vestigal side-band|
```
jacobagilbert, line 106: This is minor, but VSB can also used with digital modulations (see: ATSC)

extensions/signal.sigmf-ext.md
```diff
@@ -0,0 +1,285 @@
...
+|value|description|
+|----|-------|
+|`fhss`|frequency-hopping spread spectrum|
+|`thss`|time-hopping spread spectrum|
+|`dsss`|direct-sequence spread spectrum|
+|`css`|chirp spread spectrum|
+
+###### The `bandwidth` Field
+
+The `bandwidth` field has a float value that describes the channel bandwidth of the signal. Note that this is different from what is reported in the `core` namespace within an annotation, describing the occupied spectrum of a signal, which may or may not be comparable to the actual channel bandwidth of the communications system.
```
jacobagilbert, line 188: Can we change this name to be more specific then? `channel_bw` or `channel_bandwidth`?

extensions/signal.sigmf-ext.md
```diff
@@ -0,0 +1,285 @@
...
+
+###### The `channel` Field
+
+The `channel` field has an unsigned integer value that describes the channel number of the signal within the communication system.
+
+###### The `class_variant` Field
+
+The `class_variant` field describes any modifier to the modulation class not covered by any of the other fields. Examples include pi/4-DQPSK and GMSK.
+
+## Examples
```
jacobagilbert, line 198: Awesome providing examples!!

jacobagilbert (changes requested):
I really like this overall, very well thought out.

I do think that there is some overlap between this and some other objects and that we might be able to flatten this by one level before merging; comments are inline.

Teque5:
Suggestion for extensions in general: consider section numbering similar to [how NTIA has adopted sigmf extensions](https://github.com/NTIA/sigmf-ns-ntia/blob/master/ntia-emitter.sigmf-ext.md).

extensions/signal.sigmf-ext.md
bhilburn: This is an interesting point, @jacobagilbert. Currently, that's meant to be filled by the `comment` field, but I think what you're really highlighting is that it would be worthwhile to have a short-form human **and machine** -readable field that's in `core`. 

For purposes of process, I recommend we leave this field here, and then file a new issue & PR discussing and making that change across the codebase to keep the change atomic.

extensions/signal.sigmf-ext.md
bhilburn: Agreed, this should be made consistent with the outcomes of that discussion. As with the `label` discussion above, I'd like to /not/ hold this PR on this topic, and then once that decision is made, make that change across the code base.

extensions/signal.sigmf-ext.md
bhilburn: Interesting, I didn't know that! @Atheatos, want to make this `(analog / digital)` for completeness?

extensions/signal.sigmf-ext.md
bhilburn: Good point, and agreed with @jacobagilbert.

extensions/signal.sigmf-ext.md
bhilburn: Pulling from the examples section, this is what we have currently:

```json
{
    [...]
    "annotations": [{
        "core:sample_start": 0,
        "core:sample_count": 1000000,
        "signal:emission": {
            "label": "wifi",
            "detail": {
                "type": "digital",
                "standard": "802.11ac",
                "channel": 8
            },
            "emitter": {
                "seid": 1,
                "manufacturer": "linksys",
                "latitude": 0.000,
                "longitude": 0.000
            }
        }
    }]
}
```

What you are suggesting is change it to what would look like:

```json
{
    [...]
    "annotations": [{
        "core:sample_start": 0,
        "core:sample_count": 1000000,
        "signal:emission": {
            "label": "wifi"
        },
        "signal:detail": {
                "type": "digital",
                "standard": "802.11ac",
                "channel": 8
         },
         "signal:emitter": {
                "seid": 1,
                "manufacturer": "linksys",
                "latitude": 0.000,
                "longitude": 0.000
        }
    }]
}
```

I agree with you that it simplifies the "hierarchy" of reaching into the `emission`, but it also bloats what sits at the level of the hierarchy for the annotation, and from a semantic perspective, I actually think it makes less sense?

Not sold one way or the other. Thoughts from @n-west, @Teque5, @Atheatos?

bhilburn (changes requested):
Responded to review comments from @jacobagilbert. I think there are two small changes to make, one outstanding discussion to close (see hierarchy of `emission`), and two other items that will be addressed outside of this PR.

extensions/signal.sigmf-ext.md
jacobagilbert: This is what I am specifically proposing:

```JSON
{
    [...]
    "annotations": [{
        "core:sample_start": 0,
        "core:sample_count": 1000000,
        "core:label": "wifi",
        "signal:detail": {
                "type": "digital",
                "standard": "802.11ac",
                "channel": 8
         },
         "signal:emitter": {
                "seid": 1,
                "manufacturer": "linksys",
                "latitude": 0.000,
                "longitude": 0.000
        }
    }]
}

largely because in my use of SigMF annotations there are many times i would *only* have `label` or `signal:emitter` or `signal:detail` information to include and the extra layer of hierarchy. If we move `label` to `core` as i think we agree is a good idea, this is two functionally separate concepts (details about the emitter, and details about the emission).

extensions/signal.sigmf-ext.md
jacobagilbert: Yes, thats right. 

Are you suggesting that we merge this here then separately move it to core after this is merged?

jacobagilbert:
@Atheatos Do you have time to work on this? If not, are you OK if I hijack this to get the changes made for inclusion in v1.0?

bhilburn:
@jacobagilbert - I think you're clear to pick this up and run with it. Please do it on top of @Atheatos's commits so his contributions remain represented in-tree.

Atheatos:
@jacobagilbert @bhilburn 
Hey, sorry for the delay in responding. Indeed, I'm not able to work on this right now, so I appreciate you moving forward to get the change integrated for the v.1.0 release

jacobagilbert:
Thanks @Atheatos I just opened a new PR though. It has your commit history so your contribution will be retained.

PR #128, opened 2021-05-27
Allow Freq (but not temporally) Bounded Annotations

### LINKED ISSUE #29: Annotate frequency but not time
Annotations require bounds in time/samples and have optional bounds in frequency. Might it be useful to allow annotations with bounds in frequency but not in time?

For example, in a recording that is known to be of a transmission (or several) whose carrier frequency is not the center frequency of the recording, to mark the carrier frequency and/or band edges.

Obviously one can create an annotation whose sample range extends from the beginning to the end of the recording, but applications could benefit from knowing that it's not a thing that _happens_ to start/end there but is actually timeless.

### LINKED ISSUE #68: Compression for SigMF Archives

jacobagilbert:
Previously `sample_count` was required, now it is optional to
allow defining spectrally (but not temporally) bounded sigmf
annotations.

Fixes #29 

Signed-off-by: Jacob Gilbert <jacob.gilbert@protonmail.com>

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -399,7 +399,7 @@ annotation segment objects:
 |name|required|type|description|
 |----|--------------|-------|-----------|
 |`sample_start`|true|uint|The sample index at which this segment takes effect.|
-|`sample_count`|true|uint|The number of samples that this segment applies to. |
+|`sample_count`|false|uint|The number of samples that this segment applies to. |
 |`generator`|false|string|Human-readable name of the entity that created this annotation.|
 |`comment`|false|string|A human-readable comment.|
 |`freq_lower_edge`|false|double|The frequency (Hz) of the lower edge of the feature described by this annotation.|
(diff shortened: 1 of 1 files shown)


bhilburn:
Nice, @jacobagilbert 👍

bhilburn (approved):

gmabey:
Seems like a nice feature, and the changes clearly describe it.  I would approve it if I could.

bhilburn:
@gmabey - In the future, you are welcome to just approve via a comment 🙂

Issue #124, opened 2021-05-25
Specificity and Improvements to `captures` array


jacobagilbert:
It is unclear to me whether the captures array elements must cover the entire file duration. Currently it is allowed to define a nonzero `sample_start` for the first captures element, but it is not possible to define an end point for this capture other than to create another element. It is also not explicit within the spec if optional fields in `captures[0]` apply until `captures[1]` is defined (or _EOF_), or if they persist until explicitly redefined.

I would also propose that a mechanism to exclude portions of data from the captures array should exist but is missing. For example one use of the captures array to provide frequency information for a segment of data; if I have a device that does not retune instantly, this would look like some duration of data at frequency 1, some duration of invalid data, then data at frequency 2, etc. One could argue that annotations could be used to flag this, however it seems awkward to use annotations to convey information that is otherwise conveniently included in the captures array. The same can be true for gain or saturation metadata.

Two suggestions for which I do not have a strong opinion on the best approach:

- Create a 'data_valid' boolean entry in the captures array
- Create a 'comment' section that can be used for this purpose
- Create a `sample_end` entry that can be used to specify discontinuous chunks of the file.

I also propose the addition of an optional 'gain' field into the core `captures` definition for the same reasons frequency is included.

These are very simple PRs (some language on capture element scope, addition of a few fields) but I would like some opinions before I do that.

bhilburn:
Responding in bullet form to organize what I think are related but separate lines of discussion:

* Regarding discontinuous chunks of the file, that's the intention of the `global_index` field. Does that not provide the relevant functionality for this item?
* Excellent point re: optional fields carrying over into the next captures element. I consider this a bug and it should be addressed explicitly by adding text to the spec that says they do **NOT** carry over, IMO.

Given the two points above, it seems like the additional ask, here, would be to add an optional "num samps" sort of field to captures, is that right? I think this is a topic we've discussed at-length in some of the GRCon SigMF breakouts, and each time have arrived at the same conclusion of not doing it. What are your thoughts?

jacobagilbert:
I think discontinuities are handled effectively with the `global_index` provided they are manifested as dropped samples (i.e.: not in the sigmf-data). The issue here is that currently there is no way to specify where a segment stops other than to specify a new segment (for which 'bad' data included in the sigmf-data for whatever reason is one example of why you may want to do this), and there is no way to categorize segments other than by `frequency` or `datetime`.

[...]

PR #129, opened 2021-05-27
Explicit Scope for Captures Fields

jacobagilbert:
Per discussion in #124 this makes the scope of captures fields
explicitly defined partially resolving the issue.

Signed-off-by: Jacob Gilbert <jacob.gilbert@protonmail.com>


bhilburn (approved):
Nice quick fix 👍

PR #135, opened 2021-06-17
geolocation: update to use GeoJSON

### LINKED ISSUE #73: `lat` and `long` in Annotations

### LINKED ISSUE #84: 'altitude' field, and being able to put coordinates in global

jacobagilbert:
This closes #73 and closes #84 by identifying how altitude data should be included in recordings, and codifies the use of GeoJSON for the new `global` object `core:geolocation`.

This also marks the `core:latitude` and `core:longitude` fields in `annotations` for deprecation in a future (breaking) release.

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -85,9 +85,8 @@ interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
 JSON keywords are used as defined in [ECMA-404](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf).
 
-Augmented Backus-Naur form (ABNF) is used as defined by [RFC
-5234](https://tools.ietf.org/html/rfc5234) and updated by [RFC
-7405](https://tools.ietf.org/html/rfc7405).
+Augmented Backus-Naur form (ABNF) is used as defined by [RFC 5234](https://tools.ietf.org/html/rfc5234)
+and updated by [RFC 7405](https://tools.ietf.org/html/rfc7405).
 
 Fields defined as "human-readable", a "string", or simply as "text" shall be
 treated as plaintext where whitespace is significant, unless otherwise
@@ -233,8 +232,9 @@ The values in each name/value pair must be one of the following datatypes:
 |string|string|A string of characters, as defined by the JSON standard.|
 |boolean|boolean|Either `true` or `false`, as defined by the JSON standard.|
 |null|null|`null`, as defined by the JSON standard.|
-|array|JSON `array`|an `array` of other values, as defined by the JSON standard.|
-|object|JSON `object`|an `object` of other values, as defined by the JSON standard.|
+|array|JSON `array`|An `array` of other values, as defined by the JSON standard.|
+|object|JSON `object`|An `object` of other values, as defined by the JSON standard.|
+|GeoJSON|GeoJSON `point` object|A single GeoJSON `point` object as defined by RFC 7946.|
 
 #### Namespaces
 
@@ -257,7 +257,7 @@ belong in `core`. These canonical extension namespaces can be found in the
 namespaces may be defined by the user as needed.
 
 1. An extension namespace MUST be defined in a single file, named
-   meta-syntactically as `N.sigmf-ext.md`, where`N` is the name of the extension.
+   meta-syntactically as `N.sigmf-ext.md`, where `N` is the name of the extension.
 2. A `N.sigmf-ext.md` file MUST be a Github-Flavored Markdown file stored in UTF-8
    encoding.
 3. Extensions MUST have version numbers. It is RECOMMENDED that extensions use
@@ -275,7 +275,7 @@ This is a list of the canonical extension namespaces defined by SigMF:
 
  * `antenna` - Used to describe the antenna(s) used to for the recording.
  * `modulation` - Defines how to describe modulations used in wireless communications systems.
- * `volatile` - Allows for continously time-varying fields, such as a moving receiver or rotating antenna.
+ * `volatile` - Allows for continuously time-varying fields, such as a moving receiver or rotating antenna.
 
 #### Global Object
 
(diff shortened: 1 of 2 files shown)


Teque5 (changes requested):
This looks good, but I think we should remove the `latitude` and `longitude` keys instead of marking them as depricated.

Also remember to update the `schema.json`. I think also in `validate.py:validate_key()` you will have to add the `geometry` type.

jacobagilbert:
@Teque5 In general I agree but removing latitude and longitude keys would be a breaking change, and if I am recalling correctly (@bhilburn help me out here) we have decided that v1.0 should not break the v0.x spec (and I am aware of current software products using them). I need to update this per discussion in #73 and I'll update the other things you pointed out, thanks.

sigmf-spec.md
```diff
@@ -293,8 +294,30 @@ the `global` object:
...
 |`extensions`|false|object|A list of extensions used by this recording.|
 
+##### The `geolocation` Field
+The `core:geolocation` field in the `global` object is used to store the
+location of the recording system. This can be either a string (ISO 26324 DOI)
+to a separate recording of geolocation data or a GeoJSON (RFC 7946) geometry
+object as shown below. Point geometry objects are the only supported type for
+this field. Per the GeoJSON specification, the point coordinates are defined
+as `laitude`, `longitude` (required, in degrees), and `altitude` (optional, in
+meters above the WGS84 ellipsoid).
```
bhilburn, line 308: I think we should add detail that makes it clear that:
* those fields must appear in that order
* they are all WGS-84
* this is defined by [RFC 5870](https://datatracker.ietf.org/doc/html/rfc5870), which GeoJSON implements

Also, you've got a typo one line up for `latitude`

sigmf-spec.md
```diff
@@ -293,8 +294,30 @@ the `global` object:
 |`recorder`|false|string|The name of the software used to make this SigMF recording.|
 |`license`|false|string|A URL for the license document under which the recording is offered; when possible, use the canonical document provided by the license author, or, failing that, a well-known one.|
 |`hw`|false |string|A text description of the hardware used to make the recording.|
+|`geolocation`|false|GeoJSON or string|Location of the recording system.|
+|`hagl`|false|double|Antenna height above ground level (in meters).|
```
bhilburn, line 298: Is "HAGL" described meaningfully somewhere that we could refer to? Honest question. @dharasty?

sigmf-spec.md
```diff
@@ -293,8 +294,30 @@ the `global` object:
 |`recorder`|false|string|The name of the software used to make this SigMF recording.|
 |`license`|false|string|A URL for the license document under which the recording is offered; when possible, use the canonical document provided by the license author, or, failing that, a well-known one.|
 |`hw`|false |string|A text description of the hardware used to make the recording.|
+|`geolocation`|false|GeoJSON or string|Location of the recording system.|
```
bhilburn, line 297: I think we need to be SUPER clear that the only thing supported is a `GeoJSON **Point**`

bhilburn (changes requested):
Great PR, @jacobagilbert. I left a lot of comments / questions, but that's because this is a super complicated topic and I want to make sure we cover the edge cases everywhere we can. It is basically closing 4 years of discussion, so, not unexpected 😄

@Teque5 - I agree, and would prefer to just remove those deprecated fields entirely. The limiting thing, here, is that we do want to maintain backwards compatibility between v1.0.0 and v0.0.1, because we know v.0.0.1 is so widely used, and we don't want to break all those systems without giving them a stably versioned release.

sigmf-spec.md
jacobagilbert: "...(in meters) as measured by Ottoman calipers from the phase center of the aperture to the mean surface level of the normal square decimeter in the direction of the center of the WGS84 ellipse"

:)

sigmf-spec.md
jacobagilbert: Agree. I will fix this.

sigmf-spec.md
```diff
@@ -293,8 +294,30 @@ the `global` object:
...
+|`hagl`|false|double|Antenna height above ground level (in meters).|
 |`extensions`|false|object|A list of extensions used by this recording.|
 
+##### The `geolocation` Field
+The `core:geolocation` field in the `global` object is used to store the
+location of the recording system. This can be either a string (ISO 26324 DOI)
+to a separate recording of geolocation data or a GeoJSON (RFC 7946) geometry
```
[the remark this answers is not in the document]
jacobagilbert: Yes, this makes sense and I agree. #99 can be rebased off this (needs changes anyway).

sigmf-spec.md
```diff
@@ -293,8 +294,30 @@ the `global` object:
...
+object as shown below. Point geometry objects are the only supported type for
```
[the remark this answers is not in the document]
jacobagilbert: agree.

sigmf-spec.md
jacobagilbert: Agree (and thanks. damn gedit's lack of spell checking)

jacobagilbert:
@bhilburn I *think* I have resolved what you were asking about (ok..not HAGL i'll let someone else chime in there...). I reference the RFC texts in only one place and used the "`point` object" language anywhere it is described to make sure that is clear. If something here isn't resolved let me know.

gmabey:
@777arc are you able to review this?

sigmf-spec.md
777arc: Did we want that extended description of HAGL added to line 298?  I don't feel strongly about it either way.

777arc (approved):
LGTM, I added a note about HAGL but don't let that keep it from getting merged in

sigmf-spec.md
dharasty: > Is "HAGL" described meaningfully somewhere that we could refer to? Honest question. @dharasty?

Alas, I search high and low for something like an authoritative glossary (such as an FCC document) and -- to my amazement -- couldn't find one.

jacobagilbert: I believe that HAGL is a generally informal term, I find what is in the PR to be sufficient.

bhilburn:
@jacobagilbert - New changes look great, but merge is blocked by conflict. Can you push an update resolving the conflict? Also, since you'll be in there, can you make your reference to RFC 7946 actually link to that IETF doc?

Once merge conflict is resolved, we are good to go here 😁

jacobagilbert:
Should be G2G @bhilburn if you want to take one more quick look at the schema since I had to manually resolve that conflict that would be great.

PR #136, opened 2021-06-17
add short form field to annotation

### LINKED ISSUE #131: Add short-form human or machine readable field to `annotations`
As part of my review on #120, i noted that the `label` field is really a generally useful feature that should be in `core`. We currently have `comment`, however this is not specified as short form and may very well be a short sentence or something otherwise not machine readable.

We have the rfml.sigmf-ext but that has a number of problems for general use (mostly that if you use that extension, you are required to include an array of labels that fit into a single, fixed (for a given dataset) label hierarchy for every single annotation to remain compliant.

I propose the addition of an optional `core:label` or `core:description` field to `annotations` for a short-form human or machine readable string. The latter is consistent with the field currently in `global` and avoids conflicting with the `label` field in rfml.sigmf-ext. I also propose that there be an optional `core:included_labels` field in `global` for users to specify what labels are in the file (software utilities can easily do this automatically) to avoid requiring users to iterate over all annotations to know what is in the file.

jacobagilbert:
Per discussion in #120 and elsewhere, there is no short-form machine readable `core` field, and many situations where this is valuable.

closes #131

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -91,7 +91,9 @@ Augmented Backus-Naur form (ABNF) is used as defined by [RFC
 
 Fields defined as "human-readable", a "string", or simply as "text" shall be
 treated as plaintext where whitespace is significant, unless otherwise
-specified.
+specified. Fields defined "human/machine-readable" should be short, simple
+text strings without whitespace that are easily understood by a human and
+readily parsed by software.
 
 ## Specification
 
@@ -405,6 +407,7 @@ annotation segment objects:
 |`sample_start`|true|uint|The sample index at which this segment takes effect.|
 |`sample_count`|false|uint|The number of samples that this segment applies to. |
 |`generator`|false|string|Human-readable name of the entity that created this annotation.|
+|`label`|false|string|A short form human/machine-readable label for the annotation.|
 |`comment`|false|string|A human-readable comment.|
 |`freq_lower_edge`|false|double|The frequency (Hz) of the lower edge of the feature described by this annotation.|
 |`freq_upper_edge`|false|double|The frequency (Hz) of the upper edge of the feature described by this annotation.|
@@ -424,6 +427,11 @@ the center frequency is at baseband, the `freq_lower_edge` and `freq_upper_edge`
 fields may be relative to baseband. It is required that both `freq_lower_edge`
 and `freq_upper_edge` be provided, or neither; the use of just one field is not
 allowed.
+
+The `label` field may be used for any purpose, but it is recommended that it be
+limited to no more than 20 characters because a common use is a short form GUI
+indicator. Similarly, any user interface making use of this field should be
+capable of displaying up to 20 characters.
 
 ### Dataset Licensing
 
--- a/sigmf/schema.json
+++ b/sigmf/schema.json
@@ -96,6 +96,11 @@
                 "required": true,
                 "help": "The number of samples described by this segment"
             },
+            "core:label": {
+                "type": "string",
+                "required": false,
+                "help": "Short form human/machine readable label"
+            },
             "core:comment": {
                 "type": "string",
                 "required": false,


Teque5:
Any changes to the spec should also update the schema.

jacobagilbert:
@Teque5 I updated the spec and changed the field name to `label` instead of `description` since I agree with your other comment. I put may/should language describing the 20 character recommended limit; i don't think it makes sense to force this but I agree it is a good idea to provide some guidance for 'best' compatibility.

sigmf-spec.md
```diff
@@ -425,6 +428,11 @@ fields may be relative to baseband. It is required that both `freq_lower_edge`
 and `freq_upper_edge` be provided, or neither; the use of just one field is not
 allowed.
 
+The `label` field may be used for any prupose, but should be limited to not more
+than 20 charachters for user interface compatibility. Similarily, any user
+interface that makes use of this field should consider visibility of up to 20
+charachters.
```
Teque5, line 434: "characters"

When you change this perhaps mention the purpose of `label` is to be a short identifier for GUIs.

Teque5 (changes requested):

sigmf-spec.md
jacobagilbert: Fixed (thanks - and I had a `prupose` in there too... wow). I tweaked the language a bit to make sure your point comes across.

Teque5 (approved):

sigmf-spec.md
```diff
@@ -91,7 +91,9 @@ Augmented Backus-Naur form (ABNF) is used as defined by [RFC
 
 Fields defined as "human-readable", a "string", or simply as "text" shall be
 treated as plaintext where whitespace is significant, unless otherwise
-specified.
+specified. Fields defined "human/machine-readable" should be a short, simple
+text string without whitespace" that can be readily understood by a human and
```
bhilburn, line 95: I think we need to retain the comment that whitespace is significant, as it communicates that the expectation is that readers do not strip whitespace from `description` fields. What did you have in mind in pulling it out?

Also, you've got a stray `"` on this line 😅

sigmf-spec.md
```diff
@@ -425,6 +428,11 @@ fields may be relative to baseband. It is required that both `freq_lower_edge`
 and `freq_upper_edge` be provided, or neither; the use of just one field is not
 allowed.
 
+The `label` field may be used for any purpose, but it should be limited to fewer
+than 20 charachters as a common use is a short form indicator for GUIs. As such,
+any user interface making use of this field should be capable of displaying up
+to 20 characters.
```
bhilburn, line 434: I think we should make this "recommended" rather than a requirement, and should use RFC 2119 (per the [Conventions](https://github.com/gnuradio/SigMF/blob/master/sigmf-spec.md#conventions-used-in-this-document)). Suggestion:

```
The `label` field may be used for any purpose. It is RECOMMENDED that the label
be fewer than 20 characters to make it easily usable by GUIs.
```

bhilburn (changes requested):
Left a question re: removal of whitespace semanticism, requested change on label spec text.

Thanks for filing this, @jacobagilbert!

sigmf-spec.md
jacobagilbert: I'm not following this; I didn't remove lines 92/93 for `"human-readable", a "string", or simply as "text"` just added the statement for `"human/machine-readable"` and I do feel that we should keep this to without whitespace if it is indeed human and machine readable (can use `_` or other non-whitespace options).  Thoughts?

Definitely have an extra `"` though :).

sigmf-spec.md
jacobagilbert: Agree this is more clear than `should` - fixed.

sigmf-spec.md
bhilburn: You're right. I just misread the diff because lol 🤦‍♂️

Spec is good as-is, pending typo change 😄

PR #137, opened 2021-06-17
set pretty to true; important since it should be the default

Teque5:
The following squashed commit was approved for public release by The Aerospace Corporation on 2020-06-17. It is covered software release request #SW19-0024. Commits made by the Communications Software Implementation Department.

### Fixes
This tiny PR resolves an oversight made during our last commit by setting the default `tofile()` method to set the default keyword `pretty=true` which saves `sigmf-meta` files with indented & sorted keys instead of a long string.

### DIFF
--- a/sigmf/sigmffile.py
+++ b/sigmf/sigmffile.py
@@ -324,7 +324,7 @@ class SigMFFile(object):
         Parameters
         ----------
         filep : object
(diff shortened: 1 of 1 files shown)


jacobagilbert (approved):
Good changes.

bhilburn:
Agreed, thank you, @Teque5!

PR #138, opened 2021-06-19
feature/num_channels

### LINKED ISSUE #133: SigMF currently lacks a way to represent stereo data
Per @Teque5's note in #117, SigMF currently lacks a way to represent stereo data (e.g., data traces from an oscilloscope). This data looks like sample pairs, just like complex sample pairs, but are not actually complex data. In the o-scope file @Teque5 generated in that Issue, the pairs are defined as a complex pair to keep them aligned, but that's not semantically accurate.

I think adding the ability to support recordings of stereo real data would not be too hard of a lift from our current spec, and would be really useful (beyond just our own logo 😄).

What about just adding `stereo` as a top-level type? Modifying the ABNF from the spec gives us:

```abnf
    dataset-format = (real / complex / stereo) ((type endianness) / byte)

    real = "r"
    complex = "c"
    stereo = "s"

    type = floating-point / signed-integer / unsigned-integer
    floating-point = "f32"
    signed-integer = "i32" / "i16"
    unsigned-integer = "u32" / "u16"

    endianness = little-endian / big-endian
    little-endian = "_le"
    big-endian = "_be"

    byte = "i8" / "u8"
```

This, for example, would mean that `si16_le` indicates stereo int_16 LE data.

Thoughts?

Teque5:
@ke8ctn @jacobagilbert 

### Features 
- Closes #133 allowing multiple interleaved channels in a single dataset file.
- `core:num_channels` key to spec with relevant tests.
- Now allow reading all samples in a simpler way similar to numpy. See new example at top of `README.md`.

### Uncertainty
- In the `read_samples()` method of `SigMFFile` there is an option to read `raw_components`. I'm not sure of the intent of this option, but my interpretation is that it should also ignore the interleaving. See my updated documentation for this method.
- If this gets merged it should probably increment the spec. I didn't change it here.

_ps: This took 3 avengers movies to complete._


bhilburn (changes requested):
Overall looks good to me.

In addition to adding the field in `global`, I think there needs to be some additional description of how that field works later in the spec. Namely, under the `## Dataset Format` heading, I think we should add additional copy under this section explaining the new field:

```md
Complex samples should be interleaved, with the in-phase component first (i.e.,
`I[0]` `Q[0]` `I[1]` `Q[1]` ... `I[n]` `Q[n]`).  Also note that the `type`
applies to not only the real but also the imaginary component of a complex
sample.
```

First pass on addition:

```md
Where `core:num_channels` in the `global` object indicates that the recording 
contains more than one channel, samples from those channels should be 
interleaved in the same manner, with the same index from each channel's 
sample serially in the recording. For example, a recording with two channels 
of `i16_le` representing real-valued stereo data from an oscilloscope, and so 
here labeled `L` for `left` and `R` for `right, the data should appear 
as `L[0]` `R[0]` `L[1]` `R[1]` ... `L[n]` `R[n]`.
```

Thoughts? It feels a bit tricky to explain super clearly here in text.

Also, FWIW, you can drop the line assigning copyright to GRF in future code. You can either retain copyright as original author (or org, like Aerospace), or omit entirely. GNU Radio no longer assigns copyright to FSF.

jacobagilbert:
I agree @bhilburn - these can probably be combined into one coherent statement:

```
Complex samples should be interleaved, with the in-phase component first (i.e.,
`I[0]` `Q[0]` `I[1]` `Q[1]` ... `I[n]` `Q[n]`). When `core:num_channels` in the
`global` object indicates that the recording contains more than one channel,
samples from those channels should be interleaved in the same manner, with
the same index from each channel's sample serially in the recording. For example,
a recording with two channels of `i16_le` representing real-valued stereo data
from an oscilloscope, and so here labeled `L` for `left` and `R` for `right, the
data should appear as `L[0]` `R[0]` `L[1]` `R[1]` ... `L[n]` `R[n]`. The data type
specified by `core:data_type` applies to all channels of data.
```

Teque5:
@bhilburn @jacobagilbert I added the text mostly as you suggested. I think this is ready to merge.

I didn't see the line about assigning copyright to GRF. If you point it out I could remove it.

sigmf/sigmffile.py
```diff
@@ -277,8 +277,8 @@ def _count_samples(self):
             sample_size = self.get_sample_size()
             num_channels = self.get_num_channels()
             sample_count = file_size // sample_size // num_channels
-            if file_size % sample_size != 0:
-                warnings.warn("File '{}' does not contain an integral number of sample. It might not be valid data.".format(self.data_file))
+            if file_size % sample_count != 0:
```
ke8ctn, line 280: This seems wrong. `file_size` and `sample_size` are both in bytes, so `file_size % sample_size == 0` iff the file contains a whole number of samples (single channel), but `file_size % sample_count` doesn't work the same way. As a simple example, consider what happens when `file_size == sample_size + 1` (for `sample_size > 1`).

To generalize for multiple channels, I think the condition should just be `file_size % (sample_size * num_channels) != 0`.

Teque5:
Latest push resolves @ke8ctn's concern.

bhilburn:
Looks like I merged in the wrong order 😂

@Teque5 - Your language looks great to me. Two things preventing merge:

1) We've got conflicts on this PR now against master, and
2) The GRF copyright assignment in this branch is in `error.py`. See this changeset:

Screenshot 2021-07-14 213239

Once resolved, this is an immediate merge 🙂

Teque5:
Okay this branch is now ready to merge. It should be done **before** the feature/logo branch (#143).

Teque5:
Great. Now that this is in I think the next priority is to submit a pr to [cusignal](https://github.com/rapidsai/cusignal) and [gr-sigmf](https://github.com/skysafe/gr-sigmf) to accept `num_channels`.

PR #139, opened 2021-06-19
feature/validator

Teque5:
While doing work on #133 I saw there was a `tool.py` that was no longer working that seemed to just be for running validation against the `schema.json`. I also saw there was some defunct code in the `validator.py` doing something similar. I cleaned both these up and added a new command once SigMF is installed.

### Features
* removed defunct tool and added entry point for `sigmf_validate` script

### Examples
This is an archive but it works on anything normally openable by `sigmffile.fromfile`:
```bash
$ sigmf_validate logo.sigmf --verbose
INFO:root:Validation OK!
```
Alternatively you can see errors if things are wack. Missing required key:
```bash
ERROR:root:In Section `global', an entry is missing required key `core:version'.
```
Json doesn't decode:
```bash
ERROR:root:Expecting property name enclosed in double quotes: line 10 column 5 (char 471)
ERROR:root:Unable to decode malformed JSON.
```
or no file exists:
```bash
ERROR:root:[Errno 2] No such file or directory: '/srv/media/Unorganized/convert/sigmf_baddff.sigmf-meta'
ERROR:root:Unable to read SigMF, bad path?
```

_sleep now_


bhilburn:
Thanks, @Teque5. 

General comment: I think a lot of the code we have in this repo is largely not helpful to people because we don't really have documentation for how to use it. The primary exception to this are the two examples in the README.md, which are wildly useful, but don't cover things like the validator, here.

I don't think it makes sense to block improvements to existing code because it doesn't solve the larger problem, but I think it is worth noting for v1.0.0. @Teque5, you are by far and away the most knowledgeable person about the Python code in this repository at this point. I think it would be extremely valuable to have more built-out docs for what's here for people who want to learn from and use it.

bhilburn:
Re-reading my comment from last night, and it's not worded quite as I would have liked 😅

Anyway, what I was trying to say is that you are the most knowledgeable person, here, @Teque5 - so if you're up for trying to build out some docs to support v1.0.0 that would be awesome. If not, maybe we can recruit someone else to take it on with providing review?

PR #140, opened 2021-07-01
Refactor Modulation Extension Namespace (Take2)

jacobagilbert:
#120 needs some attention in order to get merged into this release, so this is a modification of that PR with the following changes:

- Removed boilerplate that already exists in the core spec (this is a canonical extension and it is redundant, and having it is inconsistent w/ other in-tree extensions)
- Implemented simplified structure (flattened by one level)
- Taking note of #135 this uses consistent (GeoJSON) notation for position
- Taking note of #136 this does not re-implement the `label` key
- Removes the explicit overqualified (analog) / (digital) fields where not necessary
- Uses NTIA style section numbering (good idea @Teque5)
- Examples updated per the above
- Adds power values to the `emitter` object
- Fixed a few table headings
- Some spelling fixes

@bhilburn Two items here rely on still-open PRs but both are essentially 'done' AFAIK, once those are merged this is 'atomic'.

Closes #120 

Thanks to @Atheatos for all the work to get this to where you did!


jacobagilbert:
@bhilburn Side note: I don't understand why this is showing Commits from you on Jul 20, 2018 and the original does not... I created a new branch from Atheatos's PR. Let me know if this is an issue and how it should be resolved.

bhilburn:
I don't have requested changes, here! Thanks so much for your great work on this, @Atheatos and @jacobagilbert 🙌

PR #141, opened 2021-07-01
Update the JSON schema file

jacobagilbert:
There are quite a few problems with the current JSON schema file, stuff was mixed between global/captures and a bunch of items are missing. This pulls in all of the current sigmf-spec objects that are not immediately going to change (does not have the `extensions` field under `global` or the latitude/longitude under annotations), and orders them per the spec.

### DIFF
--- a/sigmf/schema.json
+++ b/sigmf/schema.json
@@ -8,6 +8,22 @@
                 "required": true,
                 "help": "Sample data format"
             },
+            "core:sample_rate": {
+                "type": "double",
+                "required": false,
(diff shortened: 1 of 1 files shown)


bhilburn:
Check out the changeset in @Teque5's #138 - I think this will conflict with it? Happy to be wrong, here.

jacobagilbert:
Yep. They totally ninja fixed this....

bhilburn:
lol "edit: beaten by @Teque5" 😄

PR #142, opened 2021-07-03
bugfix/hash-check

Teque5:
While going over the validator I realized it was never actually checking the calculated checksum against what was in the metadata. That fixes that and adds associated tests.

## Features
* added checksum verification on file load & creation
* improved related tests
* added example to readme


bhilburn:
Good catch and nice fix, @Teque5 🙂

PR #147, opened 2021-08-13
adding spatial extension

jacobagilbert:
This extension is intended to augment the `antenna` and `multirecording` extensions by providing a canonical system for managing phase coherent metadata for spatial RF and audio applications. Currently this is primarily focused on how to associate line-of-bearing information with annotations or reference datasets. We felt this was something that was of sufficient general utility that upstreaming was a good idea. 

Due to complicated coordinate reference systems, this extension includes a figure; not sure how folks feel about that....but it is quite helpful for keeping the interaction between the cartesian antenna element definition and the spherical nature of bearing/direction data.

PR #150, opened 2021-08-16
adding spatial extension

jacobagilbert:
#147 cannot be reopened so I am recreating this here...

This extension is intended to augment the antenna and multirecording extensions by providing a canonical system for managing phase coherent metadata for spatial RF and audio applications. Currently this is primarily focused on how to associate line-of-bearing information with annotations or reference datasets. We felt this was something that was of sufficient general utility that upstreaming was a good idea.

Due to complicated coordinate reference systems, this extension includes a figure; not sure how we feel about that....but it is quite helpful for keeping the interaction between the cartesian definition of the antenna array geometry and the spherical nature of bearing/direction data. 

There is a complicated and very large trade space of how this could have been implemented, what was chosen is heavily biased toward real-world meaningful information as used in navigation, astronomy, mapping, etc, which is the primary reason why degrees were chosen for units and AZ/EL is used to contain this information as opposed to the mathematical representation of phi/theta.

If someone is aware of an appropriate ISO/RFC for the common definition of 'azimuth' and 'elevation' that would be good to know; a quick search resulted in mostly paywalled ISO specs (ISO 19101-1, 19111, and 6709 seem like they may but paywalled).

You can view the markdown file and image directly on my branch here:
https://github.com/jacobagilbert/SigMF/blob/add_spatial_ext/extensions/spatial.sigmf-ext.md


jacobagilbert:
Updated this as the first pass had multidimensional arrays which can be problematic for parsing (e.g.: in libsigmf). Also removed the obnoxious multi-type bearing object which is similarly problematic for parsing.

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,362 @@
+# The `spatial` SigMF Extension Namespace v0.0.2
+
+This document defines the `spatial` extension namespace for the Signal Metadata
+Format (SigMF) specification. This extension namespace contains objects to help
+store information about spatially diverse data, specifically information applied
+to directional antennas and multichannel phase-coherent datasets used for signal
+direction of arrival and beamforming.
+
+Multichannel datasets can be stored in SigMF Collections or as multichannel
+interleaved data depending on the application, Collections are RECOMMENDED due
```
gmabey, line 10: Suggest changing `,` to a `.`

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,362 @@
...
+![SigMF Spatial Coordinate Reference System](spatial_crs.png)
+
+**Figure 1 - SigMF Spatial Extension Coordinate Reference Systems**
+
+As shown in Figure 1, the boresight of the aperture is defined as being in the
+direction of the positive X-axis. Based on this, as an example, a horizontal
+uniform linear array aperture would be defined along the Y-axis (see examples
+section below).
+
+## 0 Datatypes
```
gmabey, line 31: I certainly am a believer in zero-indexed arrays, but I would think that `1` would be a more intuitive starting point here.  Does numbering these sections contribute to the presentation? Like, does it make it more clear in some way?

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,362 @@
...
+      "core:sample_start": 158007123,
+      "core:sample_count": 101041,
+      "core:freq_upper_edge": 165125023.0,
+      "core:freq_lower_edge": 165112401.0,
+      "spatial:signal_azimuth": 195.017
+    },
+    ...
+  ]
+}
+```
```
gmabey, line 362: the alpha =0 background of the .png seems to hinder the visibility of the axis labels for me, using a dark theme.

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,362 @@
...
+is illustrated in Figure 1 below. The ISO 80000-2:2019 compliant cartesian and
+spherical systems have a specific relationship with the reported azimuth and
+elevation, which use the conventional geospatial definitions (degrees east of
+north, and degrees above horizon respectively).
+
+![SigMF Spatial Coordinate Reference System](spatial_crs.png)
+
+**Figure 1 - SigMF Spatial Extension Coordinate Reference Systems**
+
+As shown in Figure 1, the boresight of the aperture is defined as being in the
```
gmabey, line 26: Is this a rule or a happenstance of Figure 1?

If a rule, what benefit does this rule (boresight must be the x-axis) bring, other than making it obvious which direction boresight is? Shouldn't that be explicit instead? Boresight is well defined for many array geometries, so this becomes impossible to comply with. I think the antenna geometry *description* should be independent of the array response.

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,362 @@
...
+|`point`|true|float array|meters|Cartesian Point defined by a element array of [x,y,z] referenced to the `spatial` CRS.|
+
+
+## 1 Global
+
+The `spatial` extension adds the following fields to the `global` SigMF object:
+
+|name|required|type|units|description|
+|----|--------|----|-----|-----------|
+|`num_elements`|true|int|N/A|Defines the number of phase centers / channels collected in the Collection or multichannel Dataset.|
```
gmabey, line 86: Would this be redundant to some description in the `.sigmf-collection`? Or, among the various `.sigmf-meta` files in a collection? (since it would appear in each).
In the case of a multi-channel `.sigmf-data`, it's redundant to `num_channels`, isn't it?
If you're trying to bring multi-channel collections and multi-channel `.sigmf-data` under a single umbrella, I can see some benefit in that, but there's a price for the redundancy. The data *processing* (or, handling) of those two different scenarios are *very* different (especially if you don't have consistent time stamps). That makes me wonder how valuable this common field really is.

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,362 @@
...
+|`channel_num`|true|int|N/A|The channel number, shares index with `element_geometry`.|
```
gmabey, line 87: Now, *this* field would *not* apply to multi-channel `.sigmf-data` files, agreed? I suggest a JSON object in a collection be the one and only place that `num_elements` and `channel_num` be indicated.

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,362 @@
...
+
+This document defines the `spatial` extension namespace for the Signal Metadata
+Format (SigMF) specification. This extension namespace contains objects to help
+store information about spatially diverse data, specifically information applied
+to directional antennas and multichannel phase-coherent datasets used for signal
+direction of arrival and beamforming.
+
+Multichannel datasets can be stored in SigMF Collections or as multichannel
+interleaved data depending on the application, Collections are RECOMMENDED due
+to the easier processing and better application support.
```
gmabey, line 11: Those might be advantages, but the disadvantages (no intrinsic guarantee of a common Fs, Fc, start_time among channels) are real too.

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,362 @@
...
+
+## 3 Annotations
+
+This extension adds the following optional fields to the `annotations` SigMF
+object:
+
+|name|required|type|description|
+|----|--------|----|-----------|
+|`signal_azimuth`|false|double|Azimuth in degrees east of north associated with the specific annotation.|
+|`signal_bearing`|false|`bearing`|Bearing associated with the specific annotation.|
```
gmabey, line 182: For the case of a collection, *which* of the `num_elements` `.sigmf-meta` files should `signal_{azimuth, bearing}` go in??  Please don't say "all of them" :-D

Surely this *doesn't* belong in the `.sigmf-collection` file ...

Some things are just *so* much easier in multi-channel recordings ...

extensions/spatial.sigmf-ext.md
jacobagilbert: This extension predated formal multichannel and collection support, so some of this is a bit convoluted now. It is redundant in some situations but i'm not sure how to make this better. Collections dont specify the number of channels; the length of the `core:streams` array can be a proxy but its not guaranteed that all recordings in a collection are coherent channel data streams, nor is it guaranteed they are all used.

Conditional requirement is not something SigMF supports, and Collections are the primary way this extension is used so "required" was opted for. It is redundant for multichannel, single file datasets.

Datasets with differing timestamps or offsets should not be an issue for properly functioning processing software.

extensions/spatial.sigmf-ext.md
jacobagilbert: > Now, this field would not apply to multi-channel .sigmf-data files, agreed?

yep, this doesn't make much sense for multichannel datasets.

Some place needs to map each file to aperture arrays though, and in our work we have found that contextual information is of significant value to the individual files, so this is included here. 

multichannel datasets were sort of shoehorned in to this when they came into existence (they were never really a roadmapped feature, they just sort of came about). hrmm will need to think more. same issue with conditional requirement as-is.

jacobagilbert (commented):
Thanks for the thorough review. Im going to think more about the redundancy that multichannel datasets create. I've never had them used with this but there is some convenience there that probably shouldn't be ignored.

jacobagilbert:
I should also probably mention that this is a formerly private extension that has been in active, stable use for some time. The proposal being made in this PR is not really "lets discuss this on a conceptual level", but rather "is this of sufficient common value to the world to upstream". Some tweaks can possibly be supported, specifically for currently unused features (multichannel datasets for example), but breaking changes will not be made.

gmabey:
> I should also probably mention that this is a formerly private extension that has been in active, stable use for some time. The proposal being made in this PR is not really "lets discuss this on a conceptual level", but rather "is this of sufficient common value to the world to upstream". Some tweaks can possibly be supported, specifically for currently unused features (multichannel datasets for example), but breaking changes will not be made.

Hmm -- unless 1) it forever remains a private extensions (which I think is allowed, although surely not what you were planning), 2) the public version of this which incorporates input from the community be a fork with a new name, or 3) a backwards-incompatible version 2.0 of the extension is the first one publicly released.

Again, all of these have significant drawbacks to your work, but they are hypothetically plausible, no?

Like, you *did* want feedback, right?

extensions/spatial.sigmf-ext.md
jacobagilbert: I have added clarification specifically for the case of multi-channel datasets. Even with multichannel datasets there is no guarantee that all of the data required for the spatial extension is contained in one file (e.g.: Two four channel radios that are synchronized and aligned with each other providing 4 phase-coherent channels in two separate 4 channel files), so this is only redundant in the specific (not recommended) case where all data is within one sigmf file.

extensions/spatial.sigmf-ext.md
bhilburn: Agreed 👍🏼

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,372 @@
...
+|`el_error`|false|float|degrees|Error or uncertainty in the elevation component.|
+|`range_error`|false|float|meters|Error or uncertainty in the range component.|
+
+The `az_error`, `el_error`, and `range_error` field units are degrees, but the
+exact meaning of 'error' in this context is not explicitly defined. Applications
+SHOULD specify their specific meaning, and in general this should be interpreted
+as an uncertainty range. The error fields should be SHOULD NOT be used if the
+corresponding field estimates are not included.
+
+```json
```
bhilburn, line 61: Since you've already got an image, just an idea, how hard would it be to overlay what this specific example looks like on top of that image? Not a hard req, but, may be helpful to the reader.

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,372 @@
...
+
+### 0.2 The `cartesian_point` Object
+
+The `cartesian_point` field is used to describe a single point in the Cartesian
+coordinate reference system. This object is necessary to define the phase center
+geometries for multidimensional arrays, and other cartesian locations.
+
+|name|required|type|units|description|
+|----|--------|----|-----|-----------|
+|`point`|true|float array|meters|Cartesian Point defined by a element array of [x,y,z] referenced to the `spatial` CRS.|
```
bhilburn, line 80: "float array" isn't a valid datatype, I believe?

The approach in the rest of the SigMF spec would be to call this a JSON `array`, which is a valid datatype, and then describe it as filled with floats. Check out how we did [Collections](https://github.com/gnuradio/SigMF/blob/sigmf-v1.x/sigmf-spec.md#collection-format), for example.

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,372 @@
...
+
+The `spatial` extension adds the following fields to the `global` SigMF object:
+
+|name|required|type|units|description|
+|----|--------|----|-----|-----------|
+|`num_elements`|true|int|N/A|Defines the number of phase centers / channels collected in the Collection or multichannel Dataset.|
+|`channel_num`|true|int|N/A|The channel number, shares index with `element_geometry`.|
+
+The number of elements MUST be defined here and is constant for a given
+Collection. It may be tempting to use the `core:num_channels` field however
```
bhilburn, line 93: Assuming this field remains, would it make more sense to change "num" to "index"?

Maybe we should define a new convention for SigMF where "num" is used to count things and "index" is used to refer to a specific one?

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,372 @@
...
+channel in the dataset. If all data is contained within that dataset then the
+`channel_num` field MUST be equal to zero.
+
+## 2 Captures
+
+The `spatial` extension adds the following fields to `captures` segment objects:
+
+|name|required|type|units|description|
+|----|--------|----|-----|-----------|
+|`element_geometry`|true|`cartesian_point` array|meters|Defines the relative arrangement of the array.|
```
bhilburn, line 108: "`cartesian_point` array" should be redundant if "cartesian_point" is described as an array in its datatype description``

bhilburn (changes requested):
I love the direction of this and think it will be a super useful extension. Most of my comments are relatively minor, but one could be substantial - hopefully it's not too disruptive.

Also, the image is beautiful 👍🏼

extensions/spatial.sigmf-ext.md
jacobagilbert: In addition to the idea that not all streams in a collection represent spatially associated (coherent) data, a collection should allow for multiple separately sets of coherent data. For example a collection could contain (real example BTW):
- `geolocation.sigmf-meta` (time-changing position of collector)
- `gimbal_position.sigmf-meta` (time-changing bearing of array)
- `observation_stream_ch0.sigmf-meta` (500MS/s observation data RX0)
- `observation_stream_ch1.sigmf-meta` (500MS/s observation data RX1)
- `main_beam_ch0.sigmf-meta` (30.72MS/s primary aperture data RX0.0)
- `main_beam_ch1.sigmf-meta` (30.72MS/s primary aperture data RX0.1)
...
- `main_beam_ch7.sigmf-meta` (30.72MS/s primary aperture data RX3.1)

Specifying this explicitly in the collection _might_ help in some simpler cases, but this is only redundant in simple cases, and ultimately I think that the cost of a single redundant field is negligible especially considering it contains contextually useful information even in a standalone sense.

extensions/spatial.sigmf-ext.md
jacobagilbert: See above for the concept, the following excerpt holds true here as well I think.

> ultimately I think that the cost of a single redundant field is negligible especially considering it contains contextually useful information even in a standalone sense.

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,372 @@
...
+This extension defines the following datatypes:
+
+|type|long-form name|description|
+|----|--------------|-----------|
+|bearing|signal direction bearing|Quantitative representation of a direction with optional error.|
+|cartesian_point|cartesian position|Cartesian coordinate triplet [x, y, z]|
+
+### 0.1 The `bearing` Object
+
+The `bearing` field is used to describe relative one or two dimensional signal
```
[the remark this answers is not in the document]
jacobagilbert: > I think you should call it out as a JSON object in the long-form name, and not call it a "field" in the description.

100% agree. Will fix.

extensions/spatial.sigmf-ext.md
jacobagilbert: Going to put this on the 'todo'. representing error/uncertainty can be tricky, especially in a fixed image.

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,372 @@
...
+
+## 2 Captures
+
+The `spatial` extension adds the following fields to `captures` segment objects:
+
+|name|required|type|units|description|
+|----|--------|----|-----|-----------|
+|`element_geometry`|true|`cartesian_point` array|meters|Defines the relative arrangement of the array.|
+|`aperture_bearing`|false|bearing|degrees|Bearing of aperture boresight in this segment.|
+|`emitter_bearing`|false|bearing|degrees|Bearing of signals in this segment.|
```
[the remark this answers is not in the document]
jacobagilbert: Ah true. The bearing field specifies units for the individual fields (some of which are degrees). Good catch.

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,372 @@
...
+section below).
+
+## 0 Datatypes
+
+This extension defines the following datatypes:
+
+|type|long-form name|description|
+|----|--------------|-----------|
+|bearing|signal direction bearing|Quantitative representation of a direction with optional error.|
+|cartesian_point|cartesian position|Cartesian coordinate triplet [x, y, z]|
```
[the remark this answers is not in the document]
jacobagilbert: A `cartesian_point` is also a JSON object, and not directly a multidimensional array (which present parsing and serialization complications in flatbuffers).

```json
      "spatial:element_geometry": [
        { "point": [0, 0.3,0] },
        { "point": [0, 0.1,0] },
        { "point": [0,-0.1,0] },
        { "point": [0,-0.3,0] }
      ],
```

extensions/spatial.sigmf-ext.md
jacobagilbert: Yes, I think channel_index is more appropriate.

jacobagilbert:
@bhilburn @gmabey I believe I have addressed all of your input, though there probably remains some difference of opinion on whether certain fields are redundant. Please review my latest and let me know if it makes sense.

extensions/spatial.sigmf-ext.md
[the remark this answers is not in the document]
bhilburn: Ah, I see. Good point. I still think the JSON structure for both of these should be in the description fields for the table, though 🙂

extensions/spatial.sigmf-ext.md
bhilburn: Fair 👍🏼

extensions/spatial.sigmf-ext.md
bhilburn: Good discussion, here. I'm on-board with the proposed approach, @jacobagilbert 👍🏼

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,373 @@
...
+
+Here is an example of a 4 element aperture with square geometry in the XY plane:
+
+```json
+{
+  "global": {
+    "core:datatype": "ci16_le",
+    "core:sample_rate": 40000000,
+    "antenna:gain": 6,
+    "spatial:num_elements": 6,
```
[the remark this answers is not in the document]
jacobagilbert: good catch 👍

gmabey (commented):
I don't have any more comments/complaints

jacobagilbert:
> I don't have any more comments/complaints

Ok, thanks for your input - the changes ended up being fairly subtle but you caught some important stuff with the collections vs multichannel datasets that would have made for awkward implementations.

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,385 @@
...
+section below).
+
+## 0 Datatypes
+
+This extension defines the following datatypes:
+
+|name|long-form name|description|
+|----|--------------|-----------|
+|bearing|signal direction bearing|JSON [bearing](spatial.sigmf-ext.md#01-the-bearing-object) object containing a quantitative representation of a direction with optional error fields.|
+|cartesian_point|cartesian position|JSON [cartesian_point](spatial.sigmf-ext.md#02-the-cartesian-point-object) containing a cartesian coordinate point triplet.|
```
bhilburn, line 38: I believe the description is missing the term `object` as it appears in the line above

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,385 @@
...
+
+### 0.1 The `bearing` Object
+
+A `bearing` object is used to describe relative one or two dimensional signal
+directions. The units are always in degrees; while it is legal for this field to
```
bhilburn, line 43: I tripped on this line. I believe what this text is describing are rules for all of the fields that are in units of degrees within the bearing object, but the way it's written, it makes it sound like the `bearing Object` is a `field`. Recommend clarifying that the 0-360 // +-180 rule is for all fields /within/ the bearing Object that use the unit of degrees.

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,385 @@
...
+|`el_error`|false|float|degrees|Error or uncertainty in the elevation component.|
+|`range_error`|false|float|meters|Error or uncertainty in the range component.|
+
+The `az_error`, `el_error`, and `range_error` field units are degrees, but the
+exact meaning of 'error' in this context is not explicitly defined. Applications
+SHOULD specify their specific meaning, and in general this should be interpreted
+as an uncertainty range. The error fields SHOULD NOT be included if the
+corresponding estimate fields are not present.
+
+```json
```
bhilburn, line 61: Recommend adding a note about this indicating it's an example describing some kind of bearing

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,385 @@
...
+measurement error is needed. Only one of these SHOULD be used for an individual
+annotation; if both are provided the `signal_bearing` object has priority.
+
+## 4 Collection
+
+This extension adds the following fields to the SigMF `collection` object:
+
+|name|required|type|units|description|
+|----|--------|----|-----|-----------|
+|`element_geometry`|true|array|N/A|An array of `cartesian_point` objects that specify the nominal electrical geometry of the array elements in this collection.|
```
bhilburn, line 207: This is called `aperture_geometry` earlier in the spec

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,385 @@
...
+
+The `element_geometry` field describes the phase center geometry of each element
+relative to the `spatial` coordinate reference system and MUST be specified for
+collections implementing the `spatial` extension, though the `point` may be
+`unknown`. Single channel datasets should contain a single `point` at `[0,0,0]`.
+
+## 5 Examples
```
bhilburn, line 214: I would def add an example that uses the new Collections feature!

extensions/spatial.sigmf-ext.md
```diff
@@ -0,0 +1,385 @@
...
+relative to the `aperture_bearing`. This is useful for reference data which is
+well controlled, but is not well suited for arbitrary signals or data with more
+than one emitter location.
+
+The `element_geometry` object MAY be included in each `captures` segment to
+specify phase center geometry at different frequencies. Including this in the
+`captures` segment is optional, and is only needed for multi frequency captures
+where the element phase centers also move with frequency. This object MUST be
+defined if there is not a `sigmf-collection` file that specifies the
+`aperture_geometry`. This array MUST be of length `num_elements` (if specifying
```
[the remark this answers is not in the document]
jacobagilbert: `element_geometry` is the term I am trying to use everywhere. updated.

Issue #108, opened 2019-05-17
Conversion to JSON-LD


dbouquin:
At the GNU Radio/SETI/Breakthrough hackathon we worked out major updates to the SigMF specification that include converting .sigmf-meta to [JSON-LD](https://json-ld.org/) among other changes and file additions. I'm linking to the reference model we came up with [here](https://www.lucidchart.com/invitations/accept/7ec0c9ba-ca45-40bb-9102-22f3b244ae80). 

I propose making bug fixes to the existing specification and issuing a minor release, followed by a v1 release that would include all changes discussed at the hackathon:

![bouquin_SigMF_recording](https://raw.githubusercontent.com/dbouquin/SigMF/master/images/SigMF_recording.png)

![bouquin_SigMF_spec](https://raw.githubusercontent.com/dbouquin/SigMF/master/images/SigMF_overview.png)

bhilburn:
Tagging this for v0.1.0 development.

n-west:
I'm not sure I follow what the significance of linked data is, can you help me understand it? How would an existing .sigmf-meta file change to become linked data?

gershwinlabs:
JSON-LD is neat, but it doesn't seem mainstream enough yet to satisfy design requirement 3 (minimal requirements and dependencies).  This can also be a problematic at runtime since disconnected devices wouldn't have access to the externally-linked contexts (essentially a runtime dependency).

bhilburn:
Marking this for v2.x

Sciumo:
How about now that Google is in the game?
https://mlcommons.github.io/croissant/docs/croissant-spec.html

Teque5:
It's still not clear to us how this would improve the spec. It seems to give a resource `@context` with the originating location (?) and perhaps a linked id similar to a `doi`, but it's not clear. I'd like to see some arguments in favor as well as a barebones example.

Sciumo:
It's not about the specification, per se, anyone can make a JSON-LD equivalent as I have done.
It's about about data correlation and annotation.
For IEEE 1900.8 we are looking at croissant as a means of generating annotations on existing files.
SigMF is one of many equivalent annotations on ground truth data that already exist.
The problem is one of aggregating the equivalent annotations by convention, and not by contract.

777arc:
If it's not about the SigMF specification then we should close out this issue and bring it up in a different venue.

Sciumo:
My point was that is a more general issue that subsumes SigMF.
[...]

PR #151, opened 2021-08-25
Multiple Recordings Extension

bhilburn:
This PR replaces #99 

New draft of the multi-recordings extension based on all of the discussion we have had to-date. I have a few issues with current draft, but sharing it for general comment / discussion so we can converge on the best solution.

My issues with it:

* There is duplicated information implied in the `primary`, `channel`, and `streams` fields. For example, the existence of the `streams`  array implies that the current recording is the primary.
* I'm not certain the `channel` field is still useful when paired with the `streams` array.
* Right now, linking recordings is based solely around filenames. Should we also enable referencing the sha512 hashes?

Here's my biggest concern:

The way the previous draft of this extension worked was by specifying exactly which fields in  `core` and extension namespaces could have a multirecording link. That was clunky and painful. The new draft allows you to use it in place of any non-string datatype, which I think is cleaner and realistically the right move for future extensions. That means, though, that anytime this extension is included, parsers cannot assume that fields that are normally ints or bools are actually those -- they have to be checked to see if they are strings, first. This seems painful to me. You *could* actually consider this a violation of the "don't modify `core`" rule for extensions, since it enables a new datatype for `core` fields.

I would especially love to hear thoughts on this from the broader group. As always, we're trying to figure out how to strike the right balance of not simplicity in the design and simplicity for applications working with those recordings.

Would be great to get comments / thoughts / perspectives, @jacobagilbert , @Teque5 , @storborg , @pwicks86 , @n-west , @gmabey

### DIFF
--- a/extensions/multirecordings.sigmf-ext.md
+++ b/extensions/multirecordings.sigmf-ext.md
@@ -1,0 +1,125 @@
+# The `multirecordings` SigMF Extension Namespace v0.1.0
+
+This extension makes it possible to link multiple SigMF recordings together to
+represent a single data-capture event. This is primarily used for two purposes:
+representing dynamic aspects of one SigMF recording and
+capturing multi-channel data.
+
+For recordings that digitize multiple streams of information (e.g., a
+phased-array receiver with multiple channels), this extension enables you to
+capture the multiple streams and link them such that each one can be treated as
+a separate `SigMF Recording` or they may be treated together as one.
+
+For some recordings, there are fields that are continuously variable. Examples
+include the geolocation of a radio receiver that is on a moving vehicle, the
+azimuth of a spinning antenna aperture, and the temperature readings of a
+thermal sensor. If these were stored as metadata in the "primary" signal
+recording, they would bloat and complicate the use of that recording.
+Furthermore, it would be difficult to process any aspect of the data without
+parsing all of the recorded data. This extension enables you to treat such
+fields as *simply another `SigMF Recording`*.
+
+Per the SigMF specification, all recordings must be independently evaluated and
+be able to stand-alone as compliant recordings.
+
+## 0 Datatypes
+
+| type      | long-form name        | description                                             |
+| --------- | --------------------- | ------------------------------------------------------- |
+| recording | string recording name | A string that indicates the name of a `SigMF Recording` |
+
+The `recording` datatype is the primary mechanism by which SigMF Recordings are
+linked. This datatype is the name of a SigMF Recording, **without** extensions
+(i.e., given a recording comprising `N.sigmf-meta` and `N.sigmf-data`, the
+`recording` datatype value would simply be `N`).
+
+The metadata and dataset files of the recording MUST be in the local directory
+(i.e., local to the SigMF Recording in which they are referenced, and thus in
+the same SigMF Archive).
+
+## 1 Global
+
+This extension adds the following optional fields to the `Global` object.
+
+| name          | required | type      | description                         |
+| ------------- | -------- | --------- | ----------------------------------- |
+| `primary`     | false    | recording | The primary recording this recording is linked to.|
+| `channel`     | false    | uint      | The channel index of this recording.|
+| `streams`     | false    | object    | List of SigMF recordings that represent multiple streams.|
+
+### The `primary` field
+
+The `multirecordings:primary` field in the `global` object allows the user to
+indicate another SigMF recording that acts as the 'primary' or 'parent' of this
+one, effectively providing a mechanism for bi-directional link between SigMF recordings.
+
+If omitted, it is assumed that this file is the primary.
+
+### The `channel` field
+
+In recordings with multiple channels (e.g., phased arrays), this field is used to indicate the channel number of this specific recording.
+
+### The `streams` field
+
+The `multirecordings:streams` field is a JSON array
+of `recording` strings, as defined by this extension, that indicate multiple
+streams of data that were captured as part of the same event.
+
+This field MUST only appear in the `primary` recording's metadata file.
+
+## 2 Captures
+
+This extension does not extend `captures`.
+
+## 3 Annotations
+
+This extension does not extend `annotations`.
+
+## 4 Examples
+
+
+### 4.1 Example of a 2-channel MIMO archive
+
+File: `example-0.sigmf-meta`
+```json
+{
+    "global": {
+        "core:datatype": "i16_le",
+        "core:version": "1.0.0",
+        "core:extensions" : {
+            "multirecordings": "v0.1.0"
+        },
+        "multirecordings:streams": [
+            "example-channel-0",
+            "example-channel-1",
+        ],
+        "multirecordings:channel": 0
+    },
+    "captures": [
+        {
+            "core:sample_start": 0
+        }
+    ],
+    "annotations": []
+}
+```
+
+File: `example-1.sigmf-meta`
+```json
+{
+    "global": {
+        "core:datatype": "i16_le",
+        "core:version": "1.0.0",
+        "core:extensions" : {
+            "multirecordings": "v0.1.0"
+        },
+        "multirecordings:channel": 1
+    },
+    "captures": [
+        {
+            "core:sample_start": 0
+        }
+    ],
+    "annotations": []
+}
+```
--- a/extensions/multirecordings.sigmf-ext.md
+++ b/extensions/multirecordings.sigmf-ext.md
@@ -113,7 +113,8 @@ File: `example-1.sigmf-meta`
         "core:extensions" : {
             "multirecordings": "v0.1.0"
         },
-        "multirecordings:channel": 1
+        "multirecordings:channel": 1,
+        "multirecordings:primary": "example-0"
     },
     "captures": [
         {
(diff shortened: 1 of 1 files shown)


jacobagilbert (commented):
This is an improvement, and is significantly simplified in that it does not modify the core specification, which I believe we have agreed is not something any extension should do.

My suggestions are as follows:

1. Make the `global` `primary` field required (see inline comment).
2. I agree, while this field was originally added at my suggestion I believe it largely serves as a convenient way for users to shoot themselves in the foot.
3. We can add something like the following to cover the intent of the `channel` field:

```markdown
### The `streams` field

The `multirecordings:streams` field is a JSON array
of `recording` strings, as defined by this extension, that indicate multiple
streams of data that were captured as part of the same event.

This field MUST only appear in the `primary` recording's metadata file. The
primary recording must appear first in this list, and the order of the array
can be used to infer channel order.
```

Thanks Ben!

extensions/multirecordings.sigmf-ext.md
```diff
@@ -0,0 +1,128 @@
...
+
+## 1 Global
+
+This extension adds the following optional fields to the `Global` object.
+
+| name          | required | type      | description                         |
+| ------------- | -------- | --------- | ----------------------------------- |
+| `primary`     | false    | recording | The primary recording this recording is linked to.|
+| `channel`     | false    | uint      | The channel index of this recording.|
+| `streams`     | false    | object    | List of SigMF recordings that represent multiple streams.|
```
jacobagilbert, line 50: Maybe this should be slightly reworded to be more specific:

| `streams`     | false    | object    | List of SigMF `recording` objects that represent multiple streams.|

extensions/multirecordings.sigmf-ext.md
```diff
@@ -0,0 +1,128 @@
...
+{
+    "global": {
+        "core:datatype": "i16_le",
+        "core:version": "1.0.0",
+        "core:extensions" : {
+            "multirecordings": "v0.1.0"
+        },
+        "multirecordings:streams": [
+            "example-channel-0",
+            "example-channel-1",
```
jacobagilbert, line 96: These two lines should be:

```json
            "example-0",
            "example-1"
```

bhilburn:
Been thinking about this. I think @willcode's suggestion is the right path, and it's actually the same core concept of #108 from Daina about having a top-level metadata file that describes relationships between recordings. In addition to other benefits, the fact that each individual Recording remains the same as it would have been otherwise is huge.

I'm experimenting a bit with how to make this work, working name is a "collection". Right now, the working model is that fields  from other namespaces that can be described by recordings get added to the "collections" namespace, and in the key/value pair, the value is a [base filename, dataset hash] JSON array. There are a couple of new fields that need to get added, such as "streams" for MIMO systems.

toplevel

The downside of this is that every field that can be used at the top-level must get added to the extension. BUT, if we don't do that, then we have to overwrite the definition of those fields to allow pointing to a Recording as the value for a key, which I think is worse. Better to re-define the key in a new namespace attached to a new value datatype that create confusion in the original namespace, IMO.

Posting this to gather thoughts / feedback on the approach.

mormj:
Certainly from a practical perspective, this seems much more manageable than having individual collects optionally reference each other.  Collection seems like a reasonable name.

willcode:
A parallel application is video/audio editing MLT/EDL
https://mltframework.org/docs/mltxml/
https://shotcut.org/notes/mltxml-annotations/

jacobagilbert:
I do believe allowing arbitrary type overloading will be a parsing burden (mostly in libsigmf), and I think that something like this is reasonable. Adding a file to manage simpler more common use cases (like 2 phase coherent channels) does seem a bit heavy handed, but I've been thinking about this and there are drawbacks to all approaches i can come up with and I do not find this latest suggestion is any more onerous than others. 

> leaving the individual recordings unmodified.

This is a reasonable idea, but I think we should require an unambiguous way to identify that a file is part of a collection, and to identify and access the other files in that collection given any file that is part of said collection. This could be a required (or possibly optional...) global object which defines the top level collection file (and the presence of this field indicates it is part of a collection). If there is the objective that specific files be part of multiple independent collections then this won't work.

Final thought: @willcode was your suggestion that this be implemented as part of the core spec (and not as an extension)? It sounds like that but I am not sure.

bhilburn:
> Final thought: @willcode was your suggestion that this be implemented as part of the core spec (and not as an extension)? It sounds like that but I am not sure.

I actually didn't see this in Jeff's note, but I was thinking about this this morning and I think that's exactly what we should do - especially since it preserves existing backwards compatibility. BUT, where I think this breaks is that we need the fields in their own namespace, rather than `core`, for the field collision reasons mentioned in my last message. We've been pretty firm that the main spec only defines `core`, not other namespaces, so I think ultimately it should remain an extension.

willcode:
Hmm, I don't have a real opinion about core vs ext for collections, but it's a vertical extension, where "extensions" are horizonal. So does it really fit the concept of an extension?

jacobagilbert:
I do not thing this fits the concept of an extension. We could implement something like the following structure:

```json
{
  "global":
  {
    "core:collections":
    {
      "geolocation": "geo_datafile",
      "streams": ["channel-0", "channel-1", "channel-2", "channel-3"]
    }
  },
  ...?
}
```

Of course at that point we can probably just put this directly in the `primary` data file in a collection and avoid the need for a top level file altogether. I also thing we should be explicit about what values can take on a `recording` object... there are not very many currently and there are a lot of integer/double/object types that really should never change within a sigmf dataset. 

Do you envision use of this concept be confined to `global` objects? Objects at the `captures` and `annotation` level already have a way of specifying discrete changes, though maybe it is desirable for this to cover continuously changing features within a `captures` segment or even `annotation` objects.

bhilburn:
Responding to specific points from @willcode and @jacobagilbert -

1. There are great points in this discussion about why collections shouldn't be an extension - you've definitely moved the needle of my opinion, here, but I don't think I'm totally convinced. See end for my outstanding concern.
2. I think a top-level collections file is honestly still a good idea. It makes more sense from the perspective of how recordings are currently structured, IMO. Requiring metadata that's not about a specific recording to be carried by a Recording's sigmf-meta is a little bit of an abuse of the structure, I think.
3. I agree regarding specifying each field that can be re-defined - per my comment above, otherwise, you are overwriting datatypes in existing fields defined elsewhere, which I don't think is a good approach.
4. My thinking right now is that there is nothing in `core` outside of Global that we would allow in collections, but I kind of expect there will be in user extensions. Actually, the antenna extension has `azimuth_angle` and `elevation_angle` in annotations, which are examples of things that could be defined by a collection. Thoughts?

Okay, so, let's say these are our requirements:
* Fields that can be used with collections must be specified.
* It is not defined in an extension.
* It must be contained in a separate sigmf-collection file

Thinking about it, right now we've got three top-level objects: global, captures, annotations. It sort of feels like we are describing a fourth, honestly, but one that is optional.

So, if we pull on that thread, the next question is -- how do you specify the fields in that object? Other than top-level objects, **all** other fields are namespaced, and breaking consistency there seems confusing. Also, I would argue we /have/ to namespace them, because user extensions will certainly add new fields to the collection object under their own namespace.

I think what this distills down to is that the main spec now specifies two namespaces, core and collection, and four top-level objects. It defines the fields from core that are part of collection. Users made add new fields under extension namespaces to the collection object.

Mock-up file:
`top-level.sigmf-collection`
```json
{
    "collection": {
        "collection:extensions" : {
            "antenna": "v0.1.0"
        },

        "collection:geolocation": ["latlong", "hash"],
        "collection:hagl": ["hagl", "hash"],

        "collection:azimuth": ["antenna", "hash"],

        "collection:streams": [
            ["example-channel-0", "hash"],
            ["example-channel-1", "hash"]
        ],

        // this one is from a user extension
        "userdef:somefield": ["foobar", "hash"]
    }
}
```

I think I like this... only problem I have with it is calling the top-level object "collection" and the namespace "collection" is confusing. Maybe we change the namespace defined by the primary spec to something else to be less confusing?

Otherwise, thoughts overall?

jacobagilbert:
I agree this is effectively a 4th top level object.

You raise some very good points, but I am still not sure it is necessary to specify these in a namespace other than `core`, and actually I think it may be _slightly_ simpler to do it within `core`. Keys for the specified information remain the same, and if a key exists in (and is retrieved out of) the `collection` top level object then it will be a collection type object, if it exists and is retrieved from the `global` object, it's a fixed value.

User extensions could then also define fields as valid in `collection` and/or `global` if that is appropriate just as in the base spec. The way it is above, core SigMF fields would have `collection:param`/`core:param` in collection/global respectively, and user extensions would have `extension:param` in both (correct me if thats wrong).

Another possibility for example.sigmf-collection:
```json
{
    // does not need to be in global but since its already there and the entries are not `recording` type it might make sense
    "global": {
    },
    "collection": {
        "core:geolocation": ["latlong", "hash"],
        "core:hagl": ["hagl", "hash"],

        "antenna:azimuth": ["antenna_az", "hash"],

        "core:streams": [
            ["example-channel-0", "hash"],
            ["example-channel-1", "hash"]
        ],

        // this one is from a user extension
        "extension:field": ["datafile_for_field", "hash"]
    }
}

I may be missing the point of confusion with doing something like the above, so please let me know if so.

bhilburn:
I don't think you are missing anything -- I think I'm just more worried about having the same "key" name in the "core" extension but with two different datatypes for the "value" conditionalized on whether it's in "global" or "collection".

I would _**love**_ to get inputs from other folks, especially if it's more people that disagree with me. @n-west, @storborg, @willcode, @mormj, @Teque5?

This is really, I think, the final detail before we hammer out the last draft of this, which is itself the last outstanding feature of v1.0.0 😊

bhilburn:
Continued in #157.

PR #156, opened 2021-09-15
geolocation: fix lon/lat inversion

jacobagilbert:
This is a significant oversight in how the GeoJSON spec is implemented, which require longitude first.

https://www.rfc-editor.org/rfc/rfc7946.txt see: [Page 12]

### DIFF
--- a/extensions/signal.sigmf-ext.md
+++ b/extensions/signal.sigmf-ext.md
@@ -269,7 +269,7 @@ An example describing just the ID, power, and location of an emitter:
             "power_eirp": 43.0,
             "geolocation": {
                 "type": "point",
-                "coordinates": [38.8984531, -77.0729078]
+                "coordinates": [-77.071651, 38.897397]
             }
         }
     }]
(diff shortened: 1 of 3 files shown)


jacobagilbert:
Hey @dharasty! you were helpful in helping us select GeoJSON, could you take a look at this and give a 👍 or 👎 ?

bhilburn:
Thank you, @jacobagilbert 👍

dharasty:
The fixes look correct.

In my initial example in the discussion of #73 I was VERY careful to place longitude first.  Think of GeoJSON coordinates as [x, y] which corresponds to [lon, lat] because "x" is the east-to-west ordinate -- which is longitude.

Alas, in #73, someone introduced an erroneous example with [lat, lon, alt]... and then I made propagated that error in my subsequent discussion.  I'm so sorry if I contributed to the confusion.

If it is any consolation, this is not entirely a settled matter even in the GIS community.  I'd say most computational geography APIs use the [x, y, z] -- that is to say, the [lon, lat, alt] -- convention.  But there are notable examples of software sticking with the traditional [lat, lon] order of inputs in a given API call.

PR #157, opened 2021-09-17
SigMF Collections

bhilburn:
This PR replaces #151 and implements the design discussed in that PR's thread.

Draft implementation of SigMF Collections, creating a new top-level object in a new file. This PR updates the SigMF spec with this feature, and also updates the canonical Antenna extension to use it.

### DIFF
--- a/extensions/antenna.sigmf-ext.md
+++ b/extensions/antenna.sigmf-ext.md
@@ -8,8 +8,7 @@ The following fields are specified for SigMF Recordings.
 
 ## Global
 
-The following names are specified in the `antenna` namespace and should be used in
-the `global` object:
+The following names are specified in the `antenna` namespace and should be used in the `global` object:
 
 |name|required|type|unit|description|
 |----|--------------|-------|-------|-----------|
@@ -51,4 +50,3 @@ The following fields are specificed in SigMF Collections.
 
 |`azimuth_angle`|false|float|degrees|Angle of main beam in azimuthal plane from North.|
 |`elevation_angle`|false|float|degrees|Angle of main beam in elevation plane from horizontal.|
-|`polarization`|false|float|string|E.g. `"vertical"`, `"horizontal"`, `"slant-45"`, `"left-hand circular"`, `"right-hand circular"`.|
--- a/extensions/antenna.sigmf-ext.md
+++ b/extensions/antenna.sigmf-ext.md
@@ -1,6 +1,10 @@
-# Antenna Extension v0.9.0
+# Antenna Extension v1.0.0
 
-The `antenna` namespace extension defines static antenna parameters extending `global` and dynamic antenna parameters extending `annotations`.
+The `antenna` namespace extension defines static antenna parameters extending the `global` and `annotations` objects in SigMF Recordings, and the `collection` object in a SigMF Collection.
+
+# SigMF Recordings
+
+The following fields are specified for SigMF Recordings.
 
 ## Global
 
@@ -32,6 +36,19 @@ the `annotations` object:
 
 |name|required|type|unit|description|
 |----|--------------|-------|-------|-----------|
+|`azimuth_angle`|false|float|degrees|Angle of main beam in azimuthal plane from North.|
+|`elevation_angle`|false|float|degrees|Angle of main beam in elevation plane from horizontal.|
+|`polarization`|false|float|string|E.g. `"vertical"`, `"horizontal"`, `"slant-45"`, `"left-hand circular"`, `"right-hand circular"`.|
+
+# SigMF Collections
+
+The following fields are specificed in SigMF Collections.
+
+## Collection
+
+|name|required|type|unit|description|
+|----|--------------|-------|-------|-----------|
+
 |`azimuth_angle`|false|float|degrees|Angle of main beam in azimuthal plane from North.|
 |`elevation_angle`|false|float|degrees|Angle of main beam in elevation plane from horizontal.|
 |`polarization`|false|float|string|E.g. `"vertical"`, `"horizontal"`, `"slant-45"`, `"left-hand circular"`, `"right-hand circular"`.|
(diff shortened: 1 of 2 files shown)


sigmf-spec.md
```diff
@@ -463,6 +467,54 @@ limited to no more than 20 characters because a common use is a short form GUI
...
+|`extensions`|false|object|A list of extensions used by this collection.|
+|`streams`|false|array|An ordered array of SigMF Recording Tuples, indicating multiple recorded streams of data (e.g., phased array collections).|
+
+Example `top-level.sigmf-collection` file:
+```JSON
+{
+    "collection": {
+       "core:version": "v1.0.0",
+
+        "core:extensions" : {
```
jacobagilbert, line 502: Should `extensions` be handled the same way as in `sigmf-meta` files?

sigmf-spec.md
```diff
@@ -463,6 +467,54 @@ limited to no more than 20 characters because a common use is a short form GUI
...
+The `sigmf-collection` file contains JSON-formatted metadata adhering to the SigMF schema, with one top-level object called a `collection`. The `collection` object contains name/value pairs that describe relationships between SigMF Recordings.
+
+The `collection` object points to specific recordings via a _SigMF Recording tuple_, which references the base-name of the recording and the SHA512 hash of the metadata file. Tuples may be the singular value in a key/value pair, or provided in an ordered list via a JSON array.
+
+1. The `collection` object MUST be the only top-level object in the file.
+2. The `collection` object MUST only contain SigMF key/value pairs.
+3. Keys in the `collection` object MUST use SigMF Recording Tuples to reference recordings.
+4. SigMF Recording tuples MUST take the form of `["N", "hash"]`, where `N` is the base-name of a SigMF Recording and `hash` is the SHA512 hash of the Recording's metadata file `N.sigmf-meta`.
+5. If a value contains multiple SigMF Recording tuples, they MUST appear in a JSON array.  
+6. Recordings referenced in the `collection` MUST be EITHER in the immediate local directory, or in a local SigMF Archive.
```
jacobagilbert, line 481: I interpret this two ways:

1. The `sigmf-collection` file can be in the same directory as a tar file containing the archive
2. The `sigmf-collection` file is contained within the tar file archive

If the latter (which I think you mean...) perhaps the wording here should be:

```
Recordings referenced in the `collection` MUST be EITHER in the immediate local directory, or contained within a SigMF Archive with the `collection` file.
```

extensions/antenna.sigmf-ext.md
```diff
@@ -35,3 +39,16 @@ the `annotations` object:
...
+
+# SigMF Collections
+
+The following fields are specificed in SigMF Collections.
+
+## Collection
+
+|name|required|type|unit|description|
+|----|--------------|-------|-------|-----------|
+
```
jacobagilbert, line 51: extra line

extensions/antenna.sigmf-ext.md
```diff
@@ -35,3 +39,16 @@ the `annotations` object:
...
+|`azimuth_angle`|false|float|degrees|Angle of main beam in azimuthal plane from North.|
+|`elevation_angle`|false|float|degrees|Angle of main beam in elevation plane from horizontal.|
+|`polarization`|false|float|string|E.g. `"vertical"`, `"horizontal"`, `"slant-45"`, `"left-hand circular"`, `"right-hand circular"`.|
```
jacobagilbert, line 54: not sure how i feel about `polarization` (a string) being included right now...probably needs to wait until we define how time-changing metadata values can be stored.

sigmf-spec.md
```diff
@@ -463,6 +467,54 @@ limited to no more than 20 characters because a common use is a short form GUI
...
+
+The following names are specified in the `core` namespace for use in the `collection` object.
+
+|name|required|type|description|
+|----|--------------|-------|-----------|
+|`version`|true|string|The version of the SigMF specification used to create the metadata file.|
+|`description`|false|string|A text description of the SigMF recording.|
```
jacobagilbert, line 488: `'s/recording/collection/'`

sigmf-spec.md
```diff
@@ -463,6 +467,54 @@ limited to no more than 20 characters because a common use is a short form GUI
...
+|`author`|false|string|The author's name (and optionally e-mail address) of the form "Bruce Wayne <wayne@example.com>".|
+|`collection_doi`|false|string|The registered DOI (ISO 26324) for a Collection.|
+|`license`|false|string|A URL for the license document under which the recording is offered; when possible, use the canonical document provided by the license author, or, failing that, a well-known one.|
```
jacobagilbert, line 491: Do we need to address license conflicts?

jacobagilbert (commented):
I also think we should add the following to the `global` `core` namespace:

```json
|`collection`|false|string|The name of the `collection` this file is part of (if applicable).|
```

This will allow unambiguous traversal from an arbitrary file to all other files in the same `collection`...though it does limit a file to be in a single `collection` (which I think is also a good idea). Possibly should even make this value mandatory for all `streams`?

sigmf-spec.md
bhilburn: Check out my updated text and let me know what you think.

An aside: especially with collections, seems like we should support Archives in some form that is not in a tarball

sigmf-spec.md
```diff
@@ -126,19 +127,22 @@ metadata file contains information that describes the dataset.
 6. The names of the metadata and dataset files must be identical (excepting
    their extensions).
 
+`SigMF Collection` Rules:
```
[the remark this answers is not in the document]
bhilburn: Ahh, good idea. Have a look and let me know what you think.

bhilburn:
@jacobagilbert - Updated the PR with your feedback. Regarding restricting recordings to only being part of one collection, I think we get that for "free", here, by the global `collection` object being a singular string. Thoughts?

aromanielloNTIA:
Are there currently plans to add support for SigMF Collections to the SigMF Python package?

PR #158, opened 2021-09-20
Fixing extension string to remove 'v' character

bhilburn:
Fixing a mistake that made it into the example code in the spec that was never reflected in the spec itself.

@jacobagilbert - please double-check


jacobagilbert (approved):
LGTM - needs to be fixed before release.

PR #160, opened 2021-09-20
updating schema.json

jacobagilbert:
this requires changes for the new extension arrays and collections object. also fixed minor semantic errors in spec where 'collection' should have been used.


bhilburn:
LGTM 👍

PR #177, opened 2021-09-27
fixing indentation and missing commas in schema

jacobagilbert:
also adding a closing `}` that got omitted.


jacobagilbert:
One approval, and this is breaking our current code, so I'm merging.

PR #179, opened 2021-09-27
allow annotations greater than _or equal_ to 1

jacobagilbert:
I do not see a reason we should prohibit length 1 annotations.


bhilburn:
Agreed, and there is nothing in the spec that prohibits this 👍

PR #180, opened 2021-09-28
Fix errors creating SigMFFile objects that do not have datatype set

### LINKED ISSUE #178: Creating a SigMFFile object without optional fields fails
The `global_info` parameter is optional to `SigMFFile`, but parsing will fail without this. Specifically, the `DATATYPE_KEY = "core:datatype"` field is required. While it is required for a valid SigMF metadata file, its possible that it would be added later by a user. If the global_info field is required, it should probably be required.

to reproduce:
```python
import datetime as dt
from sigmf import SigMFFile
import numpy as np

data = np.zeros(1024, dtype=np.complex64)
data.tofile('/tmp/example.sigmf-data')
meta = SigMFFile(
    data_file='/tmp/example.sigmf-data'
)
```
result:
```
  File "/opt/SigMF/sigmf/sigmffile.py", line 97, in __init__
    self._count_samples()
  File "/opt/SigMF/sigmf/sigmffile.py", line 278, in _count_samples
    sample_size = self.get_sample_size() # size of a sample in bytes
  File "/opt/SigMF/sigmf/sigmffile.py", line 261, in get_sample_size
    return dtype_info(self.get_global_field(self.DATATYPE_KEY))['sample_size']
  File "/opt/SigMF/sigmf/sigmffile.py", line 487, in dtype_info
    dtype = datatype.lower()
AttributeError: 'NoneType' object has no attribute 'lower'
```

jacobagilbert:
This prevents users from setting a datafile, which queries information that requires the datatype meta field to be set. It also generates a more useful exception when the dtype parser is run with None input.

Closes #178 

@Teque5 I solved this in a slightly different way than described in #178 because that prevents constructing a plain `SigMFFile()` object, which is reasonable and people might be doing this already. This only changes behavior that currently causes an obscure error.

### DIFF
--- a/sigmf/sigmffile.py
+++ b/sigmf/sigmffile.py
@@ -81,6 +81,9 @@ class SigMFFile():
     def __init__(self, metadata=None, data_file=None, global_info=None, skip_checksum=False):
         self.version = None
         self.schema = None
+        self.data_file = None
+        self.sample_count = 0
+
         if metadata is None:
             self._metadata = get_default_metadata(self.get_schema())
             if not self._metadata[self.GLOBAL_KEY][self.VERSION_KEY]:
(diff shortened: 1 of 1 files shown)


Teque5 (approved):
Yea looks good. I would attempt to see if this passes tests, but tests are broken for other reasons right now.

jacobagilbert:
@Teque5 i rebased this with the fixes you made; how are you running the SigMF test code? It would be good to add a test that catches this error.

jacobagilbert:
Test pass, approval. Merging.

Issue #181, opened 2021-09-28
Collection Schema Breaks Things


Teque5:
Okay so since the collection PR was merged all the tests fail since in the default `sigmffile()` we use the `schema.json` to define a set of default metadata keys based on the schema (in `get_default_metadata()`). 

We have two options to fix this:

1. Just specifically exclude the "collection" key from the schema when importing and have the schema file hold (2) separate schemas - for `sigmf-meta` and `sigmf-collection`.
2. Split the schema into `schema-meta` and `schema-collection`.

I prefer option 2 since it's simpler.

jacobagilbert:
Might make more sense to explicitly include the appropriate keys from schema.json; as this grows its possible another top level object could make it in there.

jacobagilbert:
@Teque5 i rebased this with the fixes you made; how are you running the SigMF test code? It would be good to add a test that catches this error.

Teque5:
@jacobagilbert to test just go to the root of `SigMF/` and type `pytest .` or sometimes I do `python3 setup.py test`.

PR #182, opened 2021-09-28
hotfix/split schema into two parts; tests now work again

Teque5:
This hotfix allows tests to pass correctly again.

Before this fix, the schema for the `.sigmf-meta` and `.sigmf-collection` files were being stored in the same schema which caused two issues:

1. The `required` keys in the original `schema.json` were being used as the default metadata for now SigMF files. This didn't make sense with the collection schema in there.
2. The validator would have to know which top level keys belonged to which schema.

This PR should be accompanied by another that updates the `validator.py` to separately check that `.sigmf-collection` files are valid according to that separate schema.


Teque5:
This closes #181 .

bhilburn:
There was a conflict after merging another PR. Resolved manually and pushed.

PR #193, opened 2021-10-17
clarifying sigmf compliance language

### LINKED ISSUE #189: Add a section defining "compliance"
Right now, the definition of  "compliance" is kind of spread throughout the spec. Should make a section specifically for it.

jacobagilbert:
First pass at expanding language surrounding what "SigMF compliance" means.

Closes #189

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -46,7 +46,10 @@ Copyright of contributions to SigMF are retained by their original authors. All 
         - [Annotation Segment Objects](#annotation-segment-objects)
     - [Collection Format](#collection-format)
   - [Licensing](#licensing)
-  - [SigMF Compliance by Applications](#sigmf-compliance-by-applications)
+  - [SigMF Compliance](#sigmf-compliance)
+    - [SigMF Schema Compliance](#sigmf-schema-compliance)
+    - [SigMF Dataset Compliance](#sigmf-dataset-compliance)
+    - [SigMF Application Compliance](#sigmf-applications-compliance)
   - [Citing SigMF](#citing-sigmf)
   - [Acknowledgements](#acknowledgements)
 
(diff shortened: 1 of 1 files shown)


bhilburn (commented):
This is awesome. Thank you for taking a shot at #189. After reading this, a few thoughts:

First, I think we actually have three areas of compliance:
* SigMF schema
* SigMF Recordings
* SigMF Applications

Compliance with the schema means that the metadata format complies, regardless of how it is stored. I think this is actually what is currently written as compliance for "Datasets"

Compliance as a Recording specifies what the data must look like on-disk. You could, for example, have SigMF schema compliance but not Recording compliance if your metadata is compliant but stored in a database instead of a flat file.

Finally there is application compliance.

Thoughts?

jacobagilbert:
Makes sense. Can we call it Metadata compliance for consistency?

I'll add a recording section, which I believe also requires schema/metadata compliance.

sigmf-spec.md
```diff
@@ -557,21 +559,47 @@ deciding which open license fits your needs best. Cornell University has also
...
+requirements:
+
+1. Adheres to and supports the file rules, dataset formats, `objects`,
+   `namespaces`, and `names` specified by this document.
+1. MUST contain all REQUIRED fields with the correct datatype listed the `core`
+   namespace, and any namespace listed in the `extensions` array.
+1. MUST NOT contain fields that are not outlined in the `core` or a listed
+   `extensions` namespace.
+
+Non Compliant Datasets MAY work with SigMF Compliant applications, but this is
```
jacobagilbert, line 582: @bhilburn Hrmmmm.... addressing your feedback; does this need to be changed to "non compliant recordings" [all over] then?

jacobagilbert:
@bhilburn I made some updates to this; I'm curious if this is inline with your thinking.

jacobagilbert:
Rebased and ready to merge. @bhilburn still OK w/ this?

sigmf-spec.md
```diff
@@ -689,21 +692,60 @@ deciding which open license fits your needs best. Cornell University has also
...
+In order to be 'SigMF Compliant', a schema MUST meet the following requirements:
+
+1. Adheres to and supports the metadata file naming conventions, `objects`,
+   `namespaces`, and `names` specified by this document.
+1. MUST contain all REQUIRED fields with the correct datatype listed the `core`
+   namespace, and any namespace listed in the `extensions` array.
+1. MUST NOT contain fields that are not outlined in the `core` or a listed
+   `extensions` namespace.
+
+### SigMF Recording Compliance
```
bhilburn, line 714: I think a Recording being compliant should explicitly state that the Schema must be compliant. We've already got a bullet for the Dataset

bhilburn (commented):
The primary thing I think we're missing here is Collections, which doesn't appear anywhere. Thoughts on best way to handle that?

bhilburn:
Looks great, @jacobagilbert. Love the most recent edition 👍

PR #195, opened 2021-10-30
Fix for dataset filenames containing periods

### LINKED ISSUE #194: Filenames with periods not properly supported
There is a bug affecting `sigmffile.fromfile()` and `sigmffile.SigMFFile.tofile()`. The problem occurs when attempting to load or save a SigMF dataset that has a filename containing a period, other than the delimiter between the base filename and extension, and the user provides the filename without including the extension.

For example:
> dataset = sigmf.sigmffile.fromfile('goes-16_raw_2020-09-15_220116.481975')

The issue is due to the way `sigmffile.get_sigmf_filenames()` attempts to remove extensions from filenames. I am submitting a PR with a proposed fix shortly.

ke8ctn:
This PR addresses a bug that causes filenames containing periods (other than the delimiter between the base filename and extension) to be handled incorrectly when loading or saving datasets.

The issue is due to the way `sigmffile.get_sigmf_filenames()` attempts to remove extensions from filenames, which is done to make inclusion of SigMF filename extensions optional for the user. The proposed fix is to treat anything other than valid SigMF filename extensions (`.sigmf-meta`, `.sigmf-data`, or `.sigmf`) as part of the filename when using `get_sigmf_filenames()`.

Fixes #194 .

### DIFF
--- a/sigmf/sigmffile.py
+++ b/sigmf/sigmffile.py
@@ -630,9 +630,11 @@ def get_sigmf_filenames(filename):
     Returned as dict with 'data_fn', 'meta_fn', and 'archive_fn' as keys.
 
     Keyword arguments:
-    filename -- the SigMF filename
+    filename -- the SigMF dataset filename (with or without .sigmf* extension)
     """
-    filename = path.splitext(filename)[0]
+    filename_lower = filename.lower()
+    if filename_lower.endswith(SIGMF_DATASET_EXT) or filename_lower.endswith(SIGMF_METADATA_EXT) or filename_lower.endswith(SIGMF_ARCHIVE_EXT):
+        filename = path.splitext(filename)[0]
     return {'data_fn': filename+SIGMF_DATASET_EXT, 'meta_fn': filename+SIGMF_METADATA_EXT, 'archive_fn': filename+SIGMF_ARCHIVE_EXT}
 
 


ke8ctn:
This fixes #194

sigmf/sigmffile.py
```diff
@@ -630,9 +630,11 @@ def get_sigmf_filenames(filename):
     Returned as dict with 'data_fn', 'meta_fn', and 'archive_fn' as keys.
 
     Keyword arguments:
-    filename -- the SigMF filename
+    filename -- the SigMF dataset filename (with or without .sigmf* extension)
     """
-    filename = path.splitext(filename)[0]
+    filename_lower = filename.lower()
+    if filename_lower.endswith(SIGMF_DATASET_EXT) or filename_lower.endswith(SIGMF_METADATA_EXT) or filename_lower.endswith(SIGMF_ARCHIVE_EXT):
+        filename = path.splitext(filename)[0]
```
jacobagilbert, line 637: Another possibility is that if this conditional is not met this function should error or return an empty dictionary.

sigmf/sigmffile.py
```diff
@@ -630,9 +630,11 @@ def get_sigmf_filenames(filename):
     Returned as dict with 'data_fn', 'meta_fn', and 'archive_fn' as keys.
 
     Keyword arguments:
-    filename -- the SigMF filename
+    filename -- the SigMF dataset filename (with or without .sigmf* extension)
     """
-    filename = path.splitext(filename)[0]
+    filename_lower = filename.lower()
+    if filename_lower.endswith(SIGMF_DATASET_EXT) or filename_lower.endswith(SIGMF_METADATA_EXT) or filename_lower.endswith(SIGMF_ARCHIVE_EXT):
```
jacobagilbert, line 636: `.sigmf-collection` is a valid name here also now - probably need to incorporate that into this somehow.

sigmf/sigmffile.py
```diff
@@ -630,9 +630,11 @@ def get_sigmf_filenames(filename):
     Returned as dict with 'data_fn', 'meta_fn', and 'archive_fn' as keys.
 
     Keyword arguments:
-    filename -- the SigMF filename
+    filename -- the SigMF dataset filename (with or without .sigmf* extension)
```
jacobagilbert, line 633: Its not clear to me that the filename here _should_ be supported without the .sigmf* extension. The uses in `tofile` and `fromfile` are a bit ambiguous so a resolution should clarify the permitted filenames to those functions also. I read the intention here as "needs to be a valid sigmf filename".

jacobagilbert (commented):
Hi @ke8ctn thanks for submitting this, the issue you raised and this PR have definitely uncovered some ambiguity (at best) as well as some tangentially related omissions in the `sigmf-collection` addition.

I'm not certain this is the correct resolution here, so I posted some inline comments for feedback and have tagged some other folks for review. Im interested in your feedback as well, specifically if you have reasons _why_ the tofile and fromfile should support filename arguments without sigmf extensions.

sigmf/sigmffile.py
ke8ctn: In my mind, the `SigMFFile` object is a way of interacting with the entire dataset (metadata _and_ samples), so semantically I think it makes more sense for the user to refer to a (non-archive) dataset by its base name rather than either the `.sigmf-meta` filename or `.sigmf-data` filename specifically.

bhilburn: Worth noting, here, is that once #190 gets merged, the Dataset file can be specified from within the Metadata file, and it may not share the same basename. 

From a **code** perspective, though, I believe @ke8ctn's understanding of how it's treated is correct, here. Really need @Teque5 to weigh-in, though, as the original author.

sigmf/sigmffile.py
bhilburn: That's true, but I'm not sure how it fits, here, because Collections point to `SigMFFile`s. It may the be case that this code needs to assume a unitary Recording, regardless of whether or not it's part of a collection?

bhilburn (commented):
I think we really need @Teque5 to have a look at a couple of the comments, here 🙂

sigmf/sigmffile.py
jacobagilbert: I suppose the thinking here is that we will need a way to reference the collection from the file, however in reading your comment there are two issues with my original thought, first collection names are necessarily different from sigmf files, and more importantly the collection file is already an explicit metadata key.

So marking resolved.

sigmf/sigmffile.py
ke8ctn: Oh, good point. With #190 I think it will be cleaner to require a `.sigmf-meta` or `.sigmf` filename, since we'' still be able to infer the dataset filename from the metadata (or its filename), but not the other way around. That has me leaning more toward requiring the full metadata filename with its extension.

Also, I just realized that all this time I've been saying "SigMF **dataset**" when I should have been saying "SigMF **recording**".

bhilburn:
Is this one waiting on @Teque5? Just checking in on status.

Teque5:
I don't think I have any input - I generally agree with @ke8ctn about SigMF and think we should have 1 meta file and 1 data file for each pair and KISS.

jacobagilbert:
@ke8ctn:

> That has me leaning more toward requiring the full metadata filename with its extension included.

I agree with this also. If that is the consensus, is this PR necessary? If so, can you update to indicate explicitly that the full filename is required?

bhilburn:
@jacobagilbert @ke8ctn - thoughts on what the resolution should be, here?

ke8ctn:
Since using the full metadata filename will be prescribed, and since the relevant sections of code will need to be changed significantly for #190 anyway, I'm going to go ahead and close this PR.

PR #198, opened 2021-11-29
Support metadata only distributions, complete nonstandard dataset support

### LINKED ISSUE #183: Allow for metadata-only distributions
As discussed at GRCon'21, there are many cases where you want to distribute just the Metadata file without the attached dataset. The SigMF spec should allow for this in a way that is not "spec breaking".

jacobagilbert:
This defines a new field `virtual_dataset` which can define a recording as a metadata only distribution, and explicitly states that applications are not required to support non-compliant datasets or virtual datasets (will conflict with #193, i'll rebase after that is merged).

This also improves tooling and schema for non-standard dataset files, though I am sure as we are adding these features there are other unknown complications in the way we handle nonstandard datasets...

Also fixes some weird formatting with the flowchart ​https://github.com/gnuradio/SigMF/blob/sigmf-v1.x/sigmf-spec.md#files.

Closes #183


sigmf-spec.md
```diff
@@ -336,43 +338,58 @@ about the Recording itself.
...
+| `data_doi`       | false    | string  | The registered DOI (ISO 26324) for a Recording's Dataset file.|
+| `recorder`       | false    | string  | The name of the software used to make this SigMF Recording.|
+| `license`        | false    | string  | A URL for the license document under which the Recording is offered.|
+| `hw`             | false    | string  | A text description of the hardware used to make the Recording.|
+| `dataset`        | false    | string  | The full filename of the dataset file this Metadata file describes.|
+| `trailing_bytes` | false    | uint    | The number of bytes to ignore at the end of a Non-Conforming Dataset file.|
+| `virtual_dataset`| false    | bool    | Indicates that the dataset is virtual (SigMF Recording is metadata only).|
+| `geolocation`    | false    | GeoJSON `point` object | The location of the Recording system.|
+| `extensions`     | false    | array   | A list of JSON Objects describing extensions used by this Recording.|
+| `collection`     | false    | string  | The base filename of a `collection` with which this Recording is associated.|
```
bhilburn, line 361: I can't tell what happened, here. Is there a whitespace change that got added in to this table?

sigmf-spec.md
```diff
@@ -336,43 +338,58 @@ about the Recording itself.
...
 this Metadata describes. If provided, this string MUST be the complete filename of the
 dataset file, including the extension. The dataset file must be in the local directory,
 and this string MUST NOT include any aspects of filepath other than the filename.
 
-If this field is omitted, the dataset file MUST be a SigMF Dataset file (NOT a
+If this field is omitted, the dataset MUST have a SigMF compliant Dataset file (NOT a
 Non-Conforming Dataset), and MUST have the same base filename as the Metadata file and
-use the `.sigmf-data` extension.
+use the `.sigmf-data` extension. If it is acceptable to rename the files in the SigMF
+recording, it is RECOMMENDED that SigMF Compliant recordings DO NOT use this field.
```
bhilburn, line 373: Hm, I don't think I'm following this new sentence in the spec. What are you trying to capture with this bit?

sigmf/schema-meta.json
```diff
@@ -71,6 +71,21 @@
...
+            },
+            "core:trailing_bytes": {
+                "type": "uint",
+                "required": false,
+                "help": "The number of bytes to ignore at the end of a Non-Conforming Dataset file."
+            },
+            "core:virtual_dataset": {
+                "type": "bool",
+                "required": false,
+                "help": "Always `true`, indicates that the dataset is virtual (SigMF Recording is metadata only)."
```
bhilburn, line 87: The `Always true` part of this is confusing me, I think?

sigmf-spec.md
```diff
@@ -336,43 +338,58 @@ about the Recording itself.
...
+| `author`         | false    | string  | A text identifier for the author potentially including name, handle, email, and/or other ID like Amateur Call Sign. For example "Bruce Wayne <bruce@waynetech.com>" or "Bruce (K3X)".|
+| `meta_doi`       | false    | string  | The registered DOI (ISO 26324) for a Recording's Metadata file.|
+| `data_doi`       | false    | string  | The registered DOI (ISO 26324) for a Recording's Dataset file.|
+| `recorder`       | false    | string  | The name of the software used to make this SigMF Recording.|
+| `license`        | false    | string  | A URL for the license document under which the Recording is offered.|
+| `hw`             | false    | string  | A text description of the hardware used to make the Recording.|
+| `dataset`        | false    | string  | The full filename of the dataset file this Metadata file describes.|
+| `trailing_bytes` | false    | uint    | The number of bytes to ignore at the end of a Non-Conforming Dataset file.|
+| `virtual_dataset`| false    | bool    | Indicates that the dataset is virtual (SigMF Recording is metadata only).|
```
bhilburn, line 358: If we're saying that a Recording means a compliant Schema and Dataset, and that a `virtual dataset` is non-compliant, doesn't that mean anything with a virtual dataset is actually not a Recording?

sigmf-spec.md
```diff
@@ -125,44 +125,46 @@ used to describe the relationships between multiple Recordings.
...
+             │   (optional file)  │   │
+             └──────────┬─────────┘   │
+                        │             │             ┌─────────────────────┐
+                        │             │  stored in  │                     │
+                        │ links       ├─────────────►    SigMF Archive    │
+                        │             │             │                     │
+                        │             │             │   (optional file)   │
+             ╔══════════▼══════════╗  │             └─────────────────────┘
+             ║                     ║  │
+             ║   SigMF Recording   ╟──┘
```
bhilburn, line 142: I've got another comment later about this, but I'm not sure we're being consistent here regarding whether or not a "Recording" can actually include a virtual dataset file.

bhilburn (changes requested):
Can't say I love this design, but, I also have no suggestions for what a better design might be.

That said, this approach does open a door that we don't address: what happens if virtual_dataset is set to `true`, but a file exists? Worse, if it's set to true, and `global:dataset` is populated? At the very least, we need to provide guidance for what overrides what. Thoughts?

sigmf-spec.md
jacobagilbert: basically this field provides an explicit link between metadata and dataset files, in addition to the implicit link that Compliant recordings have. If you define this field for a compliant recording's metadata, and then down the road that recording is renamed, there may now be conflict between the implicit and explicit defined metadata file.

sigmf-spec.md
jacobagilbert: yes - `virtual_dataset` is one character longer than the previous longest field so to preserve alignment these were all shifted one space.

sigmf/schema-meta.json
jacobagilbert: I think i am going to change this. basically the idea was "only use this key if this is a virtual dataset, therefore if its defined it should be true". this is unnecessarily restrictive though and when using metadata only sigmf as a generator, it may be desirable to simply redefine this field to `false` after generating the dataset.

sigmf-spec.md
jacobagilbert: I think what you are suggesting here is that "a sigmf recording is only a recording if it comprises both metadata and actual dataset files".

The interpretation i had used was "a recording is the fundamental unit of sigmf" and any complete data package (compliant sigmf-data + sigmf-meta, NCD sigmf-meta + other data, virtual with only sigmf-meta).

I generally agree that what this PR includes now may not be the best, and reads as possibly inconsistent.

sigmf-spec.md
jacobagilbert: does recording exclusively mean compliant SigMF dataset? I was of the mindset a NCD could also be called a recording, but maybe this is not correct. What i have here is just one way to structure this and I am open to suggestions.

jacobagilbert:
I also do not love this.

We have several potential issues with this and the current NCD support (which i described in a response to your inline comment about renaming files), and here for 'virtual' files. One valid use case for metadata only sigmf is the 'dataset generator' case where the metadata is distributed without a dataset (with `"core:virtual_dataset": true`) and then at some point the dataset is actually generated and its no longer virtual. There _could_ be an expectation to change this field, but i find requiring a modification (and subsequent change to the hash) to be unacceptable.

I believe in the specific example: "what happens if virtual_dataset is set to true, but a file exists" this situation is perfectly acceptable (and the apparent conflict potentially useful). It means that the metadata was originally published without the dataset, and indicates that the dataset may be locally generated. The renaming files issue i described is most definitely _not_ good or useful, but we cant save users from themselves everywhere.

I suppose the best thing to do here is define an explicit priority for the various mechanisms for identifying the dataset from a metadata file:

```
IF <metafile_basename>.sigmf-data EXISTS:
  _DATAFILE = <metafile_basename>.sigmf-data
ELIF core:virtual_dataset:
  _DATAFILE = None
ELIF core:dataset:
  _DATAFILE = core:dataset
ELSE:
  FAIL
```

jacobagilbert:
#203 is a far better solution - closing this.

PR #202, opened 2021-12-04
update language on allowed key names

jacobagilbert:
This is a convention that has been followed out of best practices and should be codified to ensure that field names can easily be mapped to variables within code or interface definition languages.

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -262,6 +262,12 @@ SigMF metadata fundamentally takes the form of key/value pairs:
 "namespace:name": value,
 ```
 
+Metadata field names in the top level `global` object, `captures` segment
+objects, or `annotations` objects MUST be of this form. All fields other than
+those at the top level which contain a `:` delimiter SHALL only use letters,
+numbers, and the `_` character; all other characters are forbidden. Field names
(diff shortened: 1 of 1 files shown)


sigmf-spec.md
```diff
@@ -262,6 +262,13 @@ SigMF metadata fundamentally takes the form of key/value pairs:
 "namespace:name": value,
 ```
 
+Metadata field keys in the top level `global` object, `captures` segment
+objects, or `annotations` objects MUST be of this form. Other than these top
+level objects which contain a `:` delimiter between the namespace and key name,
+field keys SHALL use only letters, numbers, and the `_` character, all other
+characters are forbidden. Additionally, field keys MUST NOT start with a number
+and keys MUST NOT not be C++ or python keywords.
```
bhilburn, line 270: Only request: we should specify which releases of C++ and Python we are restricting, here.

bhilburn (changes requested):

sigmf-spec.md
jacobagilbert: Per chat.gnuradio discussion will update to specify Python 3.10 and C++20 👍

PR #203, opened 2021-12-04
Support metadata only distributions, complete nonstandard dataset support (v2)

### LINKED ISSUE #183: Allow for metadata-only distributions

jacobagilbert:
This defines a new field `core:metadata_only` which allows a metadata file to be distributed by its self.

This also improves tooling and schema for non-standard recordings. There are some edge cases here that are resolved by an explicit dataset hierarchy in the case of conflicts:

1. The file named <METAFILE_BASENAME>.sigmf-meta if it exists
2. The file in the `core:dataset` field (Non-Compliant Dataset) if it exists
3. None (may be a metadata only distribution)

Alternate to #198

Closes #183

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -93,7 +93,7 @@ are capitalized after being introduced (e.g., Recording).
 ## Specification
 
 The SigMF specification fundamentally describes two types of information:
-datasets, and metadata associated with those datasets. Taken together, a dataset
+datasets, and metadata associated with those datasets. Taken together, a Dataset
 with its SigMF metadata is a SigMF `Recording`.
 
 Datasets, for purposes of this specification, are sets of digital measurements
@@ -103,10 +103,10 @@ created by digital synthesis or by an Analog-to-Digital Converter. They could
 also be geolocation coordinates from a GNSS receiver, temperature readings
 from a thermal sensor, or any other stored digital measurement information.
 
-Metadata describes the dataset with which it is associated. The metadata
-includes information meant for the human users of the dataset, such as a title
+Metadata describes the Dataset with which it is associated. The metadata
+includes information meant for the human users of the Dataset, such as a title
 and description, and information meant for computer applications (tools) that
-operate on the dataset.
+operate on the Dataset.
 
 This specification defines a schema for metadata using a `core` namespace that 
 is a reserved name and can only be defined by this specification. Other metadata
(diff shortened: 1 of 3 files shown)


sigmf-spec.md
```diff
@@ -750,9 +772,11 @@ requirements:
    definition can be in user documentation or within the code its self, though
    explicit documentation is RECOMMENDED.
 
-SigMF Compliant Applications MAY work with non-compliant Schema or Recordings,
-but such operation is NOT guaranteed or implied. Support for SigMF collections
-is OPTIONAL for SigMF compliant applications, however it is RECOMMENDED that
+Compliant applications are NOT REQUIRED to support Non-Compliant Datasets or
+Metadata Only schema files, but it is RECOMMENDED that they parse the respective
+metadata fields in the `gloabl` object to provide descriptive messages to users
```
bhilburn, line 777: typo

sigmf-spec.md
```diff
@@ -750,9 +772,11 @@ requirements:
...
+regarding why the files are not supported. Support for SigMF collections is
```
bhilburn, line 778: Collections should be cap'd

jacobagilbert:
Before this is merged, i need to go through and run the code changes; this was done open loop and while vscode is happy this needs to be tested.

sigmf-spec.md
```diff
@@ -356,26 +357,45 @@ object:
...
+
 The `core:dataset` field in the Global object is used to specify the dataset file that
 this Metadata describes. If provided, this string MUST be the complete filename of the
 dataset file, including the extension. The dataset file must be in the local directory,
 and this string MUST NOT include any aspects of filepath other than the filename.
 
-If this field is omitted, the dataset file MUST be a SigMF Dataset file (NOT a
-Non-Conforming Dataset), and MUST have the same base filename as the Metadata file and
-use the `.sigmf-data` extension.
+If a Recording does not have this field, it MUST have a compliant SigMF dataset (NOT
```
bhilburn, line 372: capital D

sigmf-spec.md
```diff
@@ -356,26 +357,45 @@ object:
...
+a Non-Conforming Dataset) which MUST use the same base filename as the Metadata file
+and use the `.sigmf-data` extension. If a SigMF Recording or archive is renamed this
```
bhilburn, line 374: capital A

sigmf-spec.md
```diff
@@ -356,26 +357,45 @@ object:
...
+field MUST also be updated, because of this it is RECOMMENDED that Compliant SigMF
+Recordings avoid use of this field. This field MAY NOT be used in conjunction the
+`core:metadata_only` field.
```
bhilburn, line 377: I think we should say what we want error handling to be if both of these fields are set. Something like:

"If the `core:metadata_only` and `core:dataset` fields are both set, then the `core:dataset` field SHOULD be ignored by applications."

Maybe this goes into application compliance section?

sigmf-spec.md
```diff
@@ -356,26 +357,45 @@ object:
 | `hw`            | false    | string  | A text description of the hardware used to make the Recording.|
 | `dataset`       | false    | string  | The full filename of the dataset file this Metadata file describes.|
 | `trailing_bytes`| false    | uint    | The number of bytes to ignore at the end of a Non-Conforming Dataset file.|
+| `metadata_only` | false    | bool    | Indicates the Metadata file is intentionally distributed without the dataset.|
```
bhilburn, line 360: capital D

bhilburn (changes requested):
minor changes only

sigmf-spec.md
jacobagilbert: so Dataset is not capitalized in this document, should it be changed to a proper noun everywhere it is used in this context? Recording, Archive, Collection are (and schema is not which i think is correct).

jacobagilbert:
@bhilburn comments addressed; i had to make a few changes to the code but it is behaving well with everything i can throw at it.

jacobagilbert:
@Teque5 if you have a few minutes to review the code here that would be appreciated

bhilburn:
LGTM 👍🏼

PR #204, opened 2021-12-16
cleanup extension files

jacobagilbert:
This includes removing ^M EOL characters (and wrecking git blame...sorry), using common language to describe arrays, and fixing a problematic object name in the signal extension (`class` is not a legal name). The last item is important to be done sooner than later and while I hate to make a breaking change, this field is problematic.


jacobagilbert:
@Atheatos - can you let me know what the impact (if anything) would be to you if we change the `signal:detail` field named `class` to `mod_class` for consistency with the requirement that field names `MUST NOT not be C++20 or Python 3.10 keywords`, and generally reduced ambiguity.

bhilburn:
This PR looks good to me. I'm ready to merge, but will give @Atheatos a couple of days to weigh-in, here.

Atheatos:
No issue on my end; should be good to merge

jacobagilbert:
Excellent - thanks for getting back @Atheatos!

Merging.

PR #206, opened 2021-12-21
tooling improvements for NCDs and captures access

### LINKED ISSUE #205: `read_samples()` can return inaccurate data
The use of fields added to support Non Conforming Datasets can cause a situation where `read_samples()` will return invalid data. Anytime `header_bytes` is nonzero, any data read will be misaligned. This specific case is relatively easy to address but when the read extends beyond a single captures segment **and** the following captures segment contains a nonzero number of `header_bytes` there is another more complicated failure mode.

This is further complicated by the fact that tooling does not exist to identify the specific file byte boundaries that contain an arbitrary captures segment.

jacobagilbert:
this adds a few new functions to better support non-compliant datasets
and more effectively read from files. the new captures segment file
reading function will work on compliant or noncompliant datasets and the
existing read_samples() will present a warning for noncompliant datasets
as it can potentially return invalid data. QA coverage for new functions
is included.

---

Other notes:
- this also adds a new error class `SigMFAccessError` which can be thrown when accessing missing but expected metadata fields, or invalid captures/sample elements. TODO: get rid of inline `assert`s used for this..
- comments out a currently failing QA test case (@Teque5 not sure why this is failing...)

Closes #205

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,11 @@ This module can be installed the typical way:
 pip install .
 ```
 
+To run the included QA tests:
+```bash
+pytest
+```
+
 ## Use Cases
 
 ### Load a SigMF archive; read all samples & metadata
--- a/sigmf/error.py
+++ b/sigmf/error.py
@@ -31,6 +31,12 @@ class SigMFValidationError(SigMFError):
     pass
 
 
+class SigMFAccessError(SigMFError):
+    """Exceptions related to accessing the contents of SigMF metadata, notably
+    when expexted fields are missing or accessing out of bounds captures."""
+    pass
+
+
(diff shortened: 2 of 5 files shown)


jacobagilbert:
Also, here is a visual I made to keep myself straight on the new QA datasets

image

bhilburn:
This one definitely needs @Teque5's eyes as a reviewer. @Teque5, any chance you can get to this before disappearing for the holidays? 😄

@jacobagilbert - the image you made and posted 👆🏼👆🏼 is really good. Let's get that added to the Wiki (probably under a new page about QA?) so it doesn't get lost 🙂

Issue #196, opened 2021-11-09
Consider Support for Double Precision Floats


jacobagilbert:
Per use case outlined here https://github.com/miek/inspectrum/pull/206 we may want to support f64 (and cf64?) aka double precision float data.

Thanks @777arc

jsallay:
I would vote for inclusion of double precision data for a variety of reasons.  First and foremost, VITA49 supports double precision data.  The more interoperability between SigMF and VITA49 the better.  As noted in the linked issue, double precision is the standard for floating point data in a lot of fields and in a lot of software.  Including double precision makes SigMF a more friendly format for beginners and increases its utility.  

A collection can contain multiple files, and it is my understanding that not all of the files need contain RF data.  For example, I could have a SigMF recording containing antenna pointing or other regularly sampled information.  Double precision could be needed for certain types of data.

jacobagilbert:
See #197 for an additional reason to include this.

jacobagilbert:
Closed by inclusion of feature in #228, thanks @gmabey!

PR #209, opened 2021-12-29
SigMFFile class is now an iterable for accessing data via memmap

gmabey:
It also now has .ndim and .shape members and has a __getitem__() method.

The purpose in this patch is to better facilitate the accessing of very large data sets through the use of `numpy.memmap()`.  The complications arise when handling types supported by SigMF but not by numpy, like fixed-precision complex samples.  The proposed solution still provides access to the data on disk through the use of `._memmap` (expert mode) but when using the bracket operator, useful datatypes are always returned.

I would like to think that this new approach could someday make the `readsamples()` method (in its current form) obsolete, but that's premature even if it were the right thing to do

Test with this code:
```python
#!/usr/bin/env python3

import json
import numpy
import sigmf
import warnings

warnings.filterwarnings("always")

meta = {
    "global": {
        "core:author": "Glen M",
        "core:datatype": "ri16_le",
        "core:num_channels": 2,
        "core:sample_rate": 48000,
        "core:version": "1.0.0"
    },
    "captures": [
        {
            "core:sample_start": 0
        }
    ] }

NUM_ROWS = 5

for dt in "ri16_le", "ci16_le", "rf32_le", "rf64_le", "cf32_le", "cf64_le":
    meta["global"]["core:datatype"] = dt
    for num_chan in 1,3:
        meta["global"]["core:num_channels"] = num_chan
        base_filename = dt + '_' + str(num_chan)
        meta_filename = base_filename + '.sigmf-meta'
        meta_fid = open(meta_filename, 'w')
        json.dump(meta, meta_fid)
        meta_fid.close()
        a = numpy.arange(NUM_ROWS * num_chan * (2 if 'c' in dt else 1))
        if 'i16' in dt:
            b = a.astype(numpy.int16)
        elif 'f32' in dt:
            b = a.astype(numpy.float32)
        elif 'f64' in dt:
            b = a.astype(numpy.float64)
        else:
            raise ValueError('whoops')

        b.tofile(base_filename + '.sigmf-data')

        f = sigmf.sigmffile.fromfile(meta_filename)
        print(f'{dt} with {num_chan} channels: .ndim = {f.ndim} .shape = {f.shape}')
        print(f._memmap[:])
        print(f[:])
        print([i for i in f])
        print()
```

### DIFF
--- a/sigmf/sigmffile.py
+++ b/sigmf/sigmffile.py
@@ -34,6 +34,17 @@ from .archive import SigMFArchive, SIGMF_DATASET_EXT, SIGMF_METADATA_EXT, SIGMF_
 from .utils import dict_merge, insert_sorted_dict_list
 from .error import SigMFFileError, SigMFAccessError
 
+class SigMFFileIterator():
+    def __init__(self, sigmf_file):
+        self.sigmf_file = sigmf_file
+        self.pos = 0
+
+    def __next__(self):
+        if self.pos >= len(self.sigmf_file):
+            raise StopIteration
+        a = self.sigmf_file[self.pos]
+        self.pos += 1
+        return a
 
 class SigMFFile():
     '''
@@ -124,6 +135,26 @@ class SigMFFile():
 
     def __repr__(self):
         return f'SigMFFile({self})'
+
+    def __len__(self):
+        return self._memmap.shape[0]
+
+    def __iter__(self):
+        return SigMFFileIterator(self)
+
+    def __getitem__(self, sli):
+        a = self._memmap[sli] # matches behavior of numpy.ndarray.__getitem__()
+        if self._return_type is not None:
+            # is_fixed_point and is_complex
+            if self._memmap.ndim == 2:
+                # num_channels==1
+                a = a[0].astype(self._return_type) + 1.j * a[1].astype(self._return_type)
+            elif self._memmap.ndim == 3:
+                # num_channels>1
+                a = a[:,0].astype(self._return_type) + 1.j * a[:,1].astype(self._return_type)
+            else:
+                raise ValueError("unhandled ndim in SigMFFile.__getitem__(); this shouldn't happen")
+        return a
 
     def _get_start_offset(self):
         """
@@ -387,6 +418,30 @@ class SigMFFile():
 
         self.data_file = data_file
         self._count_samples()
+
+        dtype = dtype_info(self.get_global_field(self.DATATYPE_KEY))
+        num_channels = self.get_num_channels()
+        self.ndim = 1 if (num_channels < 2) else 2
+        is_complex_data = dtype['is_complex']
+        is_fixedpoint_data = dtype['is_fixedpoint']
+
+        memmap_shape = (-1,)
+        if num_channels > 1:
+            memmap_shape = memmap_shape + (num_channels,)
+        if is_complex_data and is_fixedpoint_data:
+            # There is no corresponding numpy type, so we'll have to add another axis, length of 2
+            memmap_shape = memmap_shape + (2,)
+        self._return_type = dtype['memmap_convert_type']
+        #print('memmap()ing', self.get_global_field(self.DATATYPE_KEY), 'with', dtype['memmap_map_type'])
+        try:
+            self._memmap = np.memmap(self.data_file, offset=0, dtype=dtype['memmap_map_type']).reshape(memmap_shape)
+        except:  # TODO include likely exceptions here
+            warnings.warn('Failed to memory-map array from file')
+            self._memmap = None
+            self.shape = None
+        else:
+            self.shape = self._memmap.shape if (self._return_type is None) else self._memmap.shape[:-1]
+
         if skip_checksum:
             return None
         return self.calculate_hash()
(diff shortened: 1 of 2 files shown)


bhilburn:
@gmabey - This looks excellent! I'm marking this as non-required for the v1.0.0 tag, but I'm eager to get it into the v1.x branch as soon as it's reviewed and ready 🙂

@jacobagilbert @Teque5 - would love your reviews here

gmabey:
The first commit already had support in it for `complex128`, so with minimal touch-ups, this Fixes #196 so it may be of interest to @jacobagilbert @777arc @jsallay @sandersa-nist

gmabey:
I've now added another class that allows for accessing data without un-tar-ing an archive.  The intent is to address the case where an archive is so large that it is onerous/infeasible/obnoxious to unpack it.  Please consider exercising this patch with this script:
```python
#!/usr/bin/env python3

import json
import numpy
import sigmf
import warnings

warnings.filterwarnings("always")

global_info = {
        "core:author": "Glen M",
        "core:datatype": "ri16_le",
        "core:num_channels": 2,
        "core:sample_rate": 48000,
        "core:version": "1.0.0"
    }
capture_info = {
        "core:sample_start": 0
    }

NUM_ROWS = 5

for dt in "ri16_le", "ci16_le", "rf32_le", "rf64_le", "cf32_le", "cf64_le":
    global_info["core:datatype"] = dt
    for num_chan in 1,3:
        global_info["core:num_channels"] = num_chan
        base_filename = dt + '_' + str(num_chan)
        archive_filename = base_filename + '.sigmf'
        data_filename = base_filename + '.sigmf-data'

        a = numpy.arange(NUM_ROWS * num_chan * (2 if 'c' in dt else 1))
        if 'i16' in dt:
            b = a.astype(numpy.int16)
        elif 'f32' in dt:
            b = a.astype(numpy.float32)
        elif 'f64' in dt:
            b = a.astype(numpy.float64)
        else:
            raise ValueError('whoops')

        b.tofile(data_filename)
        meta = sigmf.SigMFFile(data_file=data_filename, global_info=global_info)
        meta.add_capture(0, metadata=capture_info)
        meta.tofile(archive_filename, toarchive=True)

        archi = sigmf.SigMFArchiveReader(archive_filename, skip_checksum=True)
        print(f'{dt} with {num_chan} channels: .ndim = {archi.ndim} .shape = {archi.shape}')
        print(archi.sigmffile._memmap[:])
        print(archi[:])
        print([i for i in archi])
        print()
```

sigmf/sigmf_hash.py
```diff
@@ -21,14 +21,23 @@
...
 
 
-def calculate_sha512(filename):
+def calculate_sha512(filename, offset_and_size=None):
     """
     Returns sha512 of filename
     """
     the_hash = hashlib.sha512()
     with open(filename, "rb") as handle:
-        for buff in iter(lambda: handle.read(4096), b""):
```
gmabey, line 32: I'm sad to see this concise line become so much more bloated ...

sigmf/sigmffile.py
```diff
@@ -183,7 +216,7 @@ def set_global_info(self, new_global):
         Overwrite the global info with a new dictionary.
         """
         self._validate_dict_in_section(new_global, self.GLOBAL_KEY)
-        self._metadata[self.GLOBAL_KEY] = new_global
+        self._metadata[self.GLOBAL_KEY] = new_global.copy()
```
gmabey, line 219: If `new_global` is a dict that is used to create multiple `.sigmf` files, then we probably don't want things like the `sha512` to be reflected in the dict at the level that invoked (for example) `SigMFFile.__init__()`

jacobagilbert (changes requested):
There are three mostly separate things going on here that are complicating review. I would recommend we get these split up so that they can be reviewed effectively:

1. Using numpy.memmap - this is a great idea, and I'd like to have a discussion about entirely switching to this for all dataset accesses.
2. Adding double precision float support - I think we will end up adding support but this should be in a separate PR.
3. Allowing access to compressed datafiles - also a good idea and possibly the only way we should be doing this. Please open another PR for this.

Let's get this refocused around #1 and figure out how to make that happen.

gmabey:
> 3\. compressed datafiles

Do you mean tarball'ed archives?  compression is different, right?

gmabey:
Ok, I'm somewhat of a klutz WRT git, but I think I've separated those two other commits into their own branches now.

Teque5 (commented):
I think this looks good, but there should be

1. At least a basic test added for this.
2. Increment version micro

Let me see what I can do...

Teque5:
@gmabey I submitted a PR to your `iterable` branch with my suggested improvements. After you merge that I vote this is ready. I incremented the *micro* too.

gmabey:
Slightly off topic from this PR, but shouldn't the `SigMFFile` class be named `SigMFRecording` instead?

Teque5:
> Slightly off topic from this PR, but shouldn't the `SigMFFile` class be named `SigMFRecording` instead?

You might be right about this, but it might be even more correct to call the top level object `SigMF`, so it's intuitive to just use things like `sigmf.SigMF.GLOBAL_KEY`.

Since that kind of change would break the API, we should decide how the versioning for the software is differentiated from the versioning for the spec before we do that PR.

jacobagilbert:
This would be a 2.x change.

Teque5 (approved):
@gmabey incorported my changes and I added testing, so this looks great now.

gmabey:
@jacobagilbert any chance I could get you to please review this PR? Also -- is there an open question as to whether using memmap is ok to always use?

Hmm -- looks like GitHub believes that @jacobagilbert requested changes, but I can't tell what was actually requested. Any insight into this would be appreciated.

jacobagilbert (approved):
This looks good to me @gmabey. I cannot see what changes i requested either for some reason, but its been so long I went ahead and had a complete re-review.

PR #210, opened 2022-01-07
Init version 1

gmabey:
Also, it would be very convenient to include in the release files a `.deb` created with `fpm`, like this:

```bash
fpm -s python -t deb --python-bin /usr/bin/python3 --python-package-name-prefix python3 setup.py
```

That's how I found this bug ...


gmabey:
My mistake -- this wasn't a clean PR.  Going to create a new one.

PR #222, opened 2022-01-29
Float64

gmabey:
Seems to be adequate for adding 64-bit floating point support.  Test code could easily be derived from PR comments from the `iterable` branch or those of the `untarless` branch.

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -222,7 +222,7 @@ Dataset formats MAY be added through extensions.
     complex = "c"
 
     type = floating-point / signed-integer / unsigned-integer
-    floating-point = "f32"
+    floating-point = "f32" / "f64"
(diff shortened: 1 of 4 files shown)


777arc:
Can you update the PR title now that it contains what is essentially 2 separate improvements.  Looks nice though, haven't tested it yet.

gmabey:
I apologize for the un-rebased confusion here and I don't know whether or how to make it clearer.

jacobagilbert:
I am inclined to accept this change based on:
- multiple user requests
- general utility
- no apparent reason that double precision floats should add burden

I'd like any final input this week if possible. Review on the actual MR is complicated because there are several commits in here but I will cherry pick the correct commits and close this unless there are objections.

PR #223, opened 2022-01-29
Untarless

gmabey:
I've now added another class that allows for accessing data without un-tar-ing an archive. The intent is to address the case where an archive is so large that it is onerous/infeasible/obnoxious to unpack it. Please consider exercising this patch with this script:

```python
#!/usr/bin/env python3

import json
import numpy
import sigmf
import warnings

warnings.filterwarnings("always")

global_info = {
        "core:author": "Glen M",
        "core:datatype": "ri16_le",
        "core:num_channels": 2,
        "core:sample_rate": 48000,
        "core:version": "1.0.0"
    }
capture_info = {
        "core:sample_start": 0
    }

NUM_ROWS = 5

for dt in "ri16_le", "ci16_le", "rf32_le", "rf64_le", "cf32_le", "cf64_le":
    global_info["core:datatype"] = dt
    for num_chan in 1,3:
        global_info["core:num_channels"] = num_chan
        base_filename = dt + '_' + str(num_chan)
        archive_filename = base_filename + '.sigmf'
        data_filename = base_filename + '.sigmf-data'

        a = numpy.arange(NUM_ROWS * num_chan * (2 if 'c' in dt else 1))
        if 'i16' in dt:
            b = a.astype(numpy.int16)
        elif 'f32' in dt:
            b = a.astype(numpy.float32)
        elif 'f64' in dt:
            b = a.astype(numpy.float64)
        else:
            raise ValueError('whoops')

        b.tofile(data_filename)
        meta = sigmf.SigMFFile(data_file=data_filename, global_info=global_info)
        meta.add_capture(0, metadata=capture_info)
        meta.tofile(archive_filename, toarchive=True)

        archi = sigmf.SigMFArchiveReader(archive_filename, skip_checksum=True)
        print(f'{dt} with {num_chan} channels: .ndim = {archi.ndim} .shape = {archi.shape}')
        print(archi.sigmffile._memmap[:])
        print(archi[:])
        print([i for i in archi])
        print()
```

### DIFF
--- a/sigmf/__init__.py
+++ b/sigmf/__init__.py
@@ -22,6 +22,7 @@ __version__ = '1.0.1'
 
 from .archive import SigMFArchive
 from .sigmffile import SigMFFile
+from .archivereader import SigMFArchiveReader
 
 from . import archive
 from . import error
@@ -29,3 +30,4 @@ from . import schema
 from . import sigmffile
 from . import validate
 from . import utils
+from . import archivereader
--- a/sigmf/archivereader.py
+++ b/sigmf/archivereader.py
@@ -1,0 +1,108 @@
+# Copyright 2021 GNU Radio Foundation
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""Access SigMF archives without extracting them."""
+
+import os
+import shutil
+import tarfile
+import tempfile
+
+from . import __version__  #, schema, sigmf_hash, validate
+from .sigmffile import SigMFFile
+from .archive import SigMFArchive, SIGMF_DATASET_EXT, SIGMF_METADATA_EXT, SIGMF_ARCHIVE_EXT
+from .utils import dict_merge, insert_sorted_dict_list
+from .error import SigMFFileError, SigMFAccessError
+
+
+class SigMFArchiveReader():
+    """Access data within SigMF archive `tar` in-place without extracting.
+
+    Parameters:
+
+      name      -- path to archive file to access. If file does not exist,
+                   or if `name` doesn't end in .sigmf, SigMFFileError is raised.
+    """
+    def __init__(self, name=None, skip_checksum=False, map_readonly=True, archive_buffer=None):
+        self.name = name
+        if self.name is not None:
+            if not name.endswith(SIGMF_ARCHIVE_EXT):
+                err = "archive extension != {}".format(SIGMF_ARCHIVE_EXT)
+                raise error.SigMFFileError(err)
+
+            tar_obj = tarfile.open(self.name)
+
+        elif archive_buffer is not None:
+            tar_obj = tarfile.open(fileobj=archive_buffer, mode='r:')
+
+        else:
+            raise ValueError('In sigmf.archivereader.__init__(), either `name` or `archive_buffer` must be not None')
+
+        json_contents = None
+        data_offset_size = None
+
+        for memb in tar_obj.getmembers():
+            if memb.isdir():  # memb.type == tarfile.DIRTYPE:
+                # the directory structure will be reflected in the member name
+                continue
+
+            elif memb.isfile():  # memb.type == tarfile.REGTYPE:
+                if memb.name.endswith(SIGMF_METADATA_EXT):
+                    json_contents = memb.name
+                    if data_offset_size is None:
+                        # consider a warnings.warn() here; the datafile should be earlier in the
+                        # archive than the metadata, so that updating it (like, adding an annotation)
+                        # is fast.
+                        pass
+                    with tar_obj.extractfile(memb) as memb_fid:
+                        json_contents = memb_fid.read()
+
+                elif memb.name.endswith(SIGMF_DATASET_EXT):
+                    data_offset_size = memb.offset_data, memb.size
+
+                else:
+                    print('A regular file', memb.name, 'was found but ignored in the archive')
+            else:
+                print('A member of type', memb.type, 'and name', memb.name, 'was found but not handled, just FYI.')
+
+        if data_offset_size is None:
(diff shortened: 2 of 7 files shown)


777arc:
Should we close #222 and just review this PR?

jacobagilbert:
@777arc I had asked @gmabey to split these out into separate requests since they are unrelated. Should be able to work through separating these things when its merge time.

jacobagilbert:
@gmabey in order to make this reviewable we really need to remove things that are not related to the untarless operation. Can you please cherry-pick the appropriate commits onto a clean copy of this branch and force-push?

Regarding your latest commit 8ba036b - I think the data should be mapped read only everywhere; this simplifies the API and I do not believe sigmf provides any constructs for modifying archive datafiles in place.

gmabey:
@jacobagilbert I'm going to suggest that #209 be fully integrated before reviewing/revising this one, since this builds on that.  That's on me for not making more progress in revising my PR with @Teque5 's suggestions -- I'll commit to doing that this week.

Regarding read-only, I'm definitely ok with changing the default to read-only, but that's not what you suggested.  I'm trying to objectively weigh my affinity for this feature as compared to the `batch_size` feature that I objected to.  My rationale in dissing `batch_size` was that it's not akin to anything in the numpy API, and my rationale for *liking* `map_readonly` is that it does have an analog in numpy-land, kind of.  That's because it's only part of `numpy.memmap()`, not `numpy.ndarray`.

I'm very comfortable with a parameter that toggles the readonlyness, *probably* because I've used `numpy.memmap()` a lot over the years -- not unlike @Teque5 's relationship with `batch_size` ... what do you think?

jacobagilbert:
We can rely on #209 but it will still need to be rebased / squashed.

I am not following the other discussion. Can you describe the reason `sigmf` would open the dataset as read/write?

gmabey:
Yes, I'm sure you're right that a rebase will be expedient.

Regarding writing to an archive, here's a *purely* *hypothetical* situation.  Suppose you have a RF recording of complex samples that is very large and you're working on an algorithm to detect anomalous transmissions: deviations from what your algorithm has previously determined to be a "normal state" for that frequency range.  And so to test, you want to *add* to the samples (like `+=`) some other signal (maybe, a wideband, short-duration burst) such that you have ground truth on the frequency and sample offset of the anomaly.  If you open the archive read-write, you could do this pretty painlessly.  Granted, you should update the `core:sha512` field (*if* you're using it) after doing so [and, I have a plan for doing that but it's not ready yet] but then you're golden -- no un-taring and retarding, nor running out of disk space.

A simpler case is: *add* some AWGN to increase the noise floor (thereby reducing the SNR) over a particular interval of samples.

Whadaya think?

jacobagilbert:
Both of these use cases are very real situations that are for signal processing, but not appropriate for SigMF to address much less provide tooling to support. This is something an external tool would be used for (and that tool could use `sigmf` to update checksum or metadata information).

gmabey:
Hmm I'm very surprised to hear this response, maybe we could work through that a bit.

I guess I really didn't think of this as `sigmf` doing the modifying, nor even really facilitating the modifying.  Instead, I claim that it's a builtin numpy feature, as the writing to disk would be a straight through the `memmap`.  Like, as compared to other features, it sure doesn't seem like a very "hard" thing to support.  So, WRT the "tooling to support" objection, seems like a pretty light lift to me -- or really no lift at all, since `numpy` does literally *all* of the actual work (or, what I would call "tooling").

I realize that prior to #209 the only supported way to get at the data was to literally copy every single sample out of the file *and* convert it to a [complex] floating point type.  Now, if a similarly large effort were required to write portions of an archive in place, then I would probably agree with you -- and object to the maintenance burden (especially all of those pesky `dtype` permutations).  However, as this PR builds on #209, then I think I disagree.

Instead, I would propose that the API guarantee that `sigmf` will not [directly] modify datasets, as long as `map_readonly` is asserted (which I agree needs to become the default).

And, if an external tool were to provide this support, would it not literally have to do all of the same operations that `SigMFArchiveReader` (oops that might need a name change) now does, except pass a different value to `mode`?  Seems kinda silly-redundant to me, if 99% of the support is already there ... like, how bad is it to pass around one additional boolean through 2 layers of method calls?

I suppose you might also have an objection to modifying the metadata of an archive?  I suppose that would then rule out adding an annotation?  But if we allow that, then we probably shouldn't object to facilitating the updating of other metadata, maybe even `sha512` ...  <shrug>

jacobagilbert:
There is no objection to modification of metadata, and allowing modifying metadata in place is a good idea given that's part of the job of `sigmf`.

gmabey:
I'd like to hear whether these features are (at this point) of interest.  I guess I'm surprised to realize (just now) that the fb89305ae5a9fe2bd85260857f4cded27046318b commit hasn't gone anywhere (or, attracted more interest). Furthermore, I'm confused as to why both commits are in this single PR -- and I would be happy to separate them, *if* they are of interest. My git acumen is increasing ...

sigmf/__init__.py
```diff
@@ -18,14 +18,16 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-__version__ = '1.0.1'
```
jacobagilbert, line 21: Can you revert this change please?

gmabey:
I'll definitely revert that `__version__` line, but I have found another bug related to reading from a buffer that I'll push at the same time (should be later today).  That raises in my mind the lack of unit test for this new class ... 

I see also that you're right, this change is independent of #209.

sigmf/__init__.py
gmabey: yes, done.

gmabey:
Just a bit of context regarding accessing a sigmf archive from a buffer, for the record. You see, I've got a SQL database with a *record* (haha) of type `BLOB` or `BYTEA` that *is* the actual SigMF archive. So, my app has got this byte array and wants to plot a specgram, and it seems pretty silly to hafta write it out to a temp file in order for these utils to access the data and the meta data, especially when `tarfile` and `numpy` support file-like-objects so well already.

Also, after pushing this last commit, I did some sort of within-github conflict-merge-thingy that I've never done before, but it seemed go ok and I hope that's really the case.

Note the lack of a test that exercises `SigMFArchiveReader`. I think I could in fact implement one next week, but I am also curious whether any other contributors to this project *like* this concept well enough to learn its interface through the process of writing a test.

sigmf/archivereader.py
```diff
@@ -0,0 +1,108 @@
...
+
+from . import __version__  #, schema, sigmf_hash, validate
+from .sigmffile import SigMFFile
+from .archive import SigMFArchive, SIGMF_DATASET_EXT, SIGMF_METADATA_EXT, SIGMF_ARCHIVE_EXT
+from .utils import dict_merge, insert_sorted_dict_list
+from .error import SigMFFileError, SigMFAccessError
+
+
+class SigMFArchiveReader():
+    """Access data within SigMF Archive.
```
Teque5, line 36:
```suggestion
    """Access data within SigMF archive `tar` in-place without extracting.
```

sigmf/sigmf_hash.py
```diff
@@ -21,14 +21,26 @@
...
 
 import hashlib
+import os
 
 
-def calculate_sha512(filename):
+def calculate_sha512(filename=None, fileobj=None, offset_and_size=None):
     """
-    Returns sha512 of filename
+    Returns sha512 of filename or fileobj
```
Teque5, line 29:
```suggestion
    Return sha512 of file or fileobj.
```

Teque5:
I like this idea and I think it's ready pending (2) items.

1. At least 1 test added to `tests/` mostly just consisting of your example at the top of the PR.
2. An example in the `README` as to how to use this approach.

gmabey:
Ok, I think I've addressed all of this points *except* that my test isn't working quite right. I think that someone more familiar with `pytest` could probably give me a hint to finish it up.

777arc:
@gmabey I think the solution to your pytest error is remove the line `with pytest.raises(error.SigMFFileError):` because that unit test isn't trying to test an error/invalid condition like the unit test I assume you copied that code from, so remove that `with` and unindent the rest of the lines and it should be good to go.  I'll look into setting up a pipeline to run the unit tests automatically.

jacobagilbert (approved):
Thanks @gmabey and @777arc this is a great thing to support.

Issue #174, opened 2021-09-25
Bug in Validator/Schema


Teque5:
We need a PR to fix a few small issues and the schema.

1. The validator makes a distinction between "sections" and "keys" and there really isn't any difference. As written the validator will fail on any key with a `list` type. We have `dict_list` and `double_list` now so we should fix this. Most conspicuously it prevents proper use of the `core:extensions` field.
2. When the `collections` PR was merged it left some comma errors in the schema.
3. The top-level key sorter in `sigmffile.py` should have `collection` added.

My coworker is going to write up a PR for this, since it prevents normal use of the "core:extensions" passing the `sigmf_validator`.

bhilburn:
Marking for v1.0.0

jacobagilbert:
Resolved with the merge of #224

PR #224, opened 2022-01-29
rework validate using official jsonschema

### LINKED ISSUE #176: assert meta.validate() multiple annotations bug
The assert meta.validate() module produces an "AssertionError" when I place two annotations with the same sample start. This should be allowable according to "sigmf-spec.md": 

"There is no limit to the number of annotations that can apply to the same group of samples. If two annotations have the same sample_start, there is no defined ordering between them." 

# create a capture key at time index 0
meta.add_capture(0, metadata={
    SigMFFile.FREQUENCY_KEY: 915000000,
    SigMFFile.DATETIME_KEY: dt.datetime.utcnow().isoformat()+'Z',
})

# add an annotation at sample 100 with length 200 & 10 KHz width
meta.add_annotation(100, 200, metadata = {
    SigMFFile.FLO_KEY: 914995000.0,
    SigMFFile.FHI_KEY: 915005000.0,
    SigMFFile.COMMENT_KEY: 'example annotation',
})    
    # add an annotation at sample 100 with length 200 & 10 KHz width
meta.add_annotation(100, 200, metadata = {
    SigMFFile.FLO_KEY: 914995000.0,
    SigMFFile.FHI_KEY: 915005000.0,
    SigMFFile.COMMENT_KEY: 'example annotation',
})

### LINKED ISSUE #186: validate of extensions doesn't seem to work
Either I made a mistake in implementing `core:extensions` or there's a problem validating it.  I added such an entry to `logo/sigmf_logo.sigmf-meta` 
```
{
    "global": {
        "core:author": "Kyle Logue, K6OF",
        "core:datatype": "ri16_le",
        "core:description": "The Official SigMF Logo",
        "core:num_channels": 2,
        "core:recorder": "OsciStudio & Audacity",
        "core:sample_rate": 48000,
        "core:sha512": "69893900f22de266485031b584c28fc3a0d4f361acd1d623698ed258e616e082d3d398af40d2ce805a804864cb0be631dba060f7410a27c0c2e497becdca53bf",
        "core:version": "1.0.0",
        "core:extensions": [
            {
                "name": "capture_details",
                "version": "1.0.0",
                "optional": false
            }
        ]
    },
    "captures": [
        {
            "core:sample_start": 0
        }
    ],
    "annotations": [
        {
            "core:comment": "logo warmup",
            "core:freq_lower_edge": -22000.0,
            "core:freq_upper_edge": 22000.0,
            "core:sample_count": 42000,
            "core:sample_start": 6000
        },
        {
            "core:comment": "logo spinup",
            "core:freq_lower_edge": -22000.0,
            "core:freq_upper_edge": 22000.0,
            "core:sample_count": 138000,
            "core:sample_start": 48000
        },
        {
            "core:comment": "logo steady",
            "core:freq_lower_edge": -22000.0,
            "core:freq_upper_edge": 22000.0,
            "core:sample_count": 96000,
            "core:sample_start": 186000
        }
    ]
}
```
and this is what I see:
```
[root@6c37442078e7 s]# /usr/local/python3_embedded/bin/sigmf_validate logo/sigmf_logo.sigmf-meta 
  File "/usr/local/python3_embedded/bin/sigmf_validate", line 33, in <module>
    sys.exit(load_entry_point('SigMF==0.0.2', 'console_scripts', 'sigmf_validate')())
  File "/usr/local/python3_embedded/lib/python3.9/site-packages/sigmf/validate.py", line 195, in main
    result = signal.validate()
  File "/usr/local/python3_embedded/lib/python3.9/site-packages/sigmf/sigmffile.py", line 320, in validate
    return validate.validate(
  File "/usr/local/python3_embedded/lib/python3.9/site-packages/sigmf/validate.py", line 154, in validate
    for result in (validate_section(data.get(key), ref.get(key), key) for key in ref):
  File "/usr/local/python3_embedded/lib/python3.9/site-packages/sigmf/validate.py", line 154, in <genexpr>
    for result in (validate_section(data.get(key), ref.get(key), key) for key in ref):
  File "/usr/local/python3_embedded/lib/python3.9/site-packages/sigmf/validate.py", line 145, in validate_section
    return {
  File "/usr/local/python3_embedded/lib/python3.9/site-packages/sigmf/validate.py", line 108, in validate_section_dict
    for result in key_validation_results:
  File "/usr/local/python3_embedded/lib/python3.9/site-packages/sigmf/validate.py", line 102, in <genexpr>
    validate_key(
  File "/usr/local/python3_embedded/lib/python3.9/site-packages/sigmf/validate.py", line 78, in validate_key
    if 'type' in ref_dict and not match_type(data_value, ref_dict["type"]):
  File "/usr/local/python3_embedded/lib/python3.9/site-packages/sigmf/validate.py", line 46, in match_type
    return value is None or {
KeyError: 'dict_list'
```

Teque5:
Okay, this was a tough one. Definitely closes #186, #174. I think this also closes #176.

# rework validate using official jsonschema
## Changes
* Remove concept of `sections` throughout.
* Removed custom json validation framework, switch to official `jsonschema`
* Rewrite meta schema.
* `validate()` is run on file write automatically
* Improve validation testing.

## API Change
* SigMFFile.validate() no longer returns `True` or `ValidationError`. Will return `None` or raise `ValidationError`

This commit increments `minor` version to `1.1.0`.

## Issues
* Unable to get defaults working in schema definition, but when that occurs `get_default_metadata` for SigMFFile should be created dynamically.
*  There are no tests for the collection schema.
*  The schema can be improved dramatically.

I was using [jsonschema.net](https://www.jsonschema.net/home) to help figure out the schemas, but a lot of the free software seems to be of varying quality. I'd love it if someone else gave this a once-over. This [schema linter](https://www.json-schema-linter.com/) was also useful in debugging.


gmabey:
@Teque5 First of all, thank you for what looks like a pretty big lift today!

I looked at every line of the diff, but didn't understand the context for much of it.  I'm commenting because I have been wondering whether a significant tools bump like this would deserve a minor level version increment or a patch level version increment (the 2nd or the 3rd number).

I see that you've moved the version to 1.1.0, and I have a suggestion for the group that I'm not 100% sold on myself, but is worth discussing.

The thought is to only perform minor-level version increments for backwards-compatible changes to the *spec*, and to have all tooling-related changes result in only a patch-level version increment.

This approach puts the emphasis of the version squarely on the spec, making the tools feel less important, but they will get over it.  This improved validation support *is* significant, but it's an incremental improvement for support of the *spec*.

Of course, major-level version increments would then only happen when some backwards-incompatible change is made to the spec.  If we want to differentiate between tools improvements and backwards-compatible spec improvements, then that probably falls to patch-level and minor-level, respectively.

Another approach would be to increment a x.y.z.w 4-number version scheme, but that's probably not the most attractive.

What do you think?

Teque5:
@gmabey I really don't think we should deviate from the standard `major.minor.micro` versioning and be something special. Usually if there is an API change you increment minor, but since this version is more related to the spec I can change it to `micro` if others agree. That would make this `1.0.2` if it's the next merged PR.

Teque5:
@gmabey and if there are places in the code you are wondering about, just comment with a `?` or something and I'll do a quick summary. I hadn't looked at the validator code in a while, but I just spent 12 hrs staring at it so I know how everything works right now.

jacobagilbert:
@Teque5 thanks for getting this together. I'll take a more in depth look and do a proper review this week.

Regarding versioning, I agree there is no reason to overcomplicate this. The right way to address specification vs tooling versioning differences is to break them out into separate efforts, and we can do so if we eventually feel it necessary.

Because this changes the API it will need to wait for the next `minor` version most likely, though we could simply retain the `True`/`False` behavior and preserve the API for now and make the bulk of the change to 1.0.2 which is probably my preference though its not super important at the moment and I have noted "versioning" as something to discuss on the next working group call.

Can you elaborate more on why you made the API change?

Teque5:
> Can you elaborate more on why you made the API change?

The `validate()` functionality was originally written for v0.0.1 and was doing some un-pythonic things. Specifically it would either return `True` or a special error object called `ValidationResult` which always required special handling for functions calling `validate()`. As implemented now it will always return `None` and will throw an exception related to jsonschema if there is an issue. If desired, the user can then use `try...except` to catch the error if they need to. I prefer this more pythonic implementation.

Teque5:
I've rebased the PR to v1.x and if someone reviews it @gmabey we can merge. This would increment version to v1.1.0 since the API changes for validation.

jacobagilbert:
Closes #236

PR #227, opened 2022-02-10
Refer to pip in README

tammojan:
I think the installation instructions should not advise everyone to build the package from source. Using pip is common practice for any python package. Of course the build from source should be in the readme, but as option b.


jacobagilbert (approved):
Agree, thanks @tammojan

Teque5 (approved):
LGTM

PR #228, opened 2022-02-15
datatype updates: add double (float64) and fix numpy bit-width error

### LINKED ISSUE #226: Validation Error with Complex64
When creating a file with:

...
SigMFFile.DATATYPE_KEY: get_data_type_str(data)
...

and running file.validate(), the returned datatype is cf64_le which fails validation as being an unsupported type.

Apologies in advance if this is a user error. I'm new here.

jacobagilbert:
Includes relevant commit from #222 (thanks @gmabey) with bug fix.

Fixes issue with converting numpy objects to sigmf datatype names identified in #226 and missing imports in the README example. also allows this function to generate datatype strings for complex integer values.

closes #222
closes #226

### DIFF
--- a/sigmf/utils.py
+++ b/sigmf/utils.py
@@ -93,7 +93,10 @@ def get_endian_str(ray):
 
 def get_data_type_str(ray):
     """
-    Return the SigMF datatype string for the datatype of `ray`.
+    Return the SigMF datatype string for the datatype of numpy array `ray`.
+
+    NOTE: this function only supports native numpy types so interleaved complex
+    integer types are not supported.
     """
     if not isinstance(ray, np.ndarray):
         raise error.SigMFError('Argument must be a numpy array')
@@ -104,10 +107,15 @@ def get_data_type_str(ray):
     data_type_str = ''
     if atype.kind == 'c':
         data_type_str += 'cf'
+        # units are component bits, numpy complex types len(I)+len(Q)
+        data_type_str += str(atype.itemsize * 8 // 2)
     elif atype.kind == 'f':
-        data_type_str += 'f'
+        data_type_str += 'rf'
+        data_type_str += str(atype.itemsize * 8)  # units are bits
     elif atype.kind in ('u','i'):
-        data_type_str += atype.kind
-    data_type_str += str(atype.itemsize * 8)  # units is bits
-    data_type_str += get_endian_str(ray)
+        data_type_str += 'r' + atype.kind
+        data_type_str += str(atype.itemsize * 8)  # units are bits
+    if (atype.itemsize > 1):
+        # only append endianness for types over 8 bits
+        data_type_str += get_endian_str(ray)
     return data_type_str
--- a/README.md
+++ b/README.md
@@ -154,6 +154,8 @@ for adx, annotation in enumerate(annotations):
 
 ```python
 import datetime as dt
+import numpy as np
+import sigmf
 from sigmf import SigMFFile
 from sigmf.utils import get_data_type_str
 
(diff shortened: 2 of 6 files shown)


sigmf/utils.py
```diff
@@ -91,9 +91,13 @@ def get_endian_str(ray):
         # endianness must be either '=' (native) or '|' (doesn't matter)
         return '_le' if sys.byteorder == 'little' else '_be'
 
-def get_data_type_str(ray):
+def get_data_type_str(ray, interleaved_complex_int = False):
```
Teque5, line 94: Not sure I'm in love with this kwarg, but I don't see an obvious alternative. Also there shouldn't be spaces around your `=`.

Teque5 (approved):
99% great.

sigmf/utils.py
jacobagilbert: I'll fix the whitespace thanks. I'm also not in love with this.... I debated leaving this out all together however after internal debate I think `ci16` is common enough to warrant it. Open to removal if there are strong opinions.

gmabey:
It was in order to support `ci16` that I originally separated out `get_endian_str()` into its own function -- so that I could automate (in separate code) the (not very hard, granted) generation of at least *that* part of the datatype string.

So, full empathy in the importance of this special case.

However, I think that the concept of `interleaved_complex_int` runs contrary to the spirit of `get_dtype_str()`.  Regardless of how you want to *store* the data, you *don't* have a numpy ndarray instance (`ray`) that is of type `ci16`, ya just don't.

sigmf/utils.py
Teque5: Yea we have to support `ci16`, but in my mind the whole point of calling the `get` is to resolve the ambiguity with the datatype. If it's not possible to determine with numpy, perhaps it should return some ambiguity like a tuple of possibilities or an error saying it cannot determine if it's `ci16` for `f8` or something.

jacobagilbert: @gmabey the issue comes about because the canonical numpy representation of ci16 data *is* as a np.int16 array.

gmabey: @jacobagilbert <sigh> maybe I should back up.  I work with `ci16` data every day, just like you.  That's why the `__getitem__()` method in #209 works with this datatype.

I think that have already dealt (very much) with the issues you and @Teque5 are discussing.  The difference is that I'm not just unhappy with `interleaved_complex_int`, I'm trying to say that it seems to be an attempt to turn `get_data_type_str()` into a universal datatype string generator, which was not my intent.

I feel that returning a tuple of possibly-compatible-but-not-equivalent-to-ray.dtype strings is an awful complication to what *was* supposed to be a straightforward *convenience* function.

Also, because I first created a new comment associated with the PR instead of replying to this one (like I should have done, sorry -- and I didn't flag you) you might have missed part of this conversation (maybe).

jacobagilbert: I also do not think ambiguous returns are what we want.

The `get_data_type_str()` function is being interpreted as a datatype string generator because the docstring says

> Return the SigMF datatype string for the datatype of numpy array `ray`.

One option is to update the docstring to indicate that this function does not work for ci16_* data, but something needs to address this part of the problem while we fix the other bug in this. In general, convenience functions should not increase edge cases, which leads me to the `cplx_interleaved_int` arg... still leaning that direction of the various not-so-god options we have.

jacobagilbert:
As no consensus was reached, this has been rolled back to support the minimal changes that fix the bugs and add f64 support.

jacobagilbert:
I will merge this next week pending any objections @gmabey @Teque5 @bhilburn

sigmf/utils.py
gmabey: Sorry for not following back up with this sooner, but I think that docstring is perfectly accurate as-is, since it's not possible to have a `numpy` array of type `ci16`.  So, anything that you *can* pass to `get_data_type_str()` will result in the correct SigMF typestring being returned.  But all that aside -- thank you for being willing to accept the lowest common denominator of consensus.

PR #233, opened 2022-03-19
remove rfml.sigmf-ext.md from canonical extensions

jacobagilbert:
This extension has no examples and it is not clear how to use it. Removal would **NOT** affect the validity of any SigMF using it, but I have been asked about what SigMF using this should look like several times over the past 6 months, and I have no useful answer.

Happy to change this PR to "include examples and documentation for rfml.sigmf-ext.md if that is possible! The 'label heirarchy' seems to be the unique feature this extension offers over what is already in the core spec.


bhilburn:
This extension was designed and used by NSWC Crane to create the datasets for DARPA's RFMLS program. Unfortunately, the government's lawyers determined those datasets could never be released, and no publicly releasable datasets were made. I'm not aware of anyone else using this particular extension as a result.

I am also in favor of removal, and if it needs to come back, more details and examples would make it more broadly useful.

jacobagilbert:
Ok, thanks for the context. I'll talk w/ Crane and see if I can get some info. If a good faith effort here fails I'll press along with deprecation on the grounds of this being proprietary.

Teque5:
I think this PR is ready.

PR #234, opened 2022-03-23
Add `presentation` Extension

jacobagilbert:
This extension was written to support suggestive input as to how SigMF Recordings and Metadata should be presented visually to users. A basic example of how this could be used is here: ~https://github.com/miek/inspectrum/pull/195~ https://github.com/miek/inspectrum/pull/208 which is just a quick proof of concept for how this might look.

There is no expectation that applications support all features of this, and the "simple" usage where an optional `presentation:color` field is added to annotations allows for the features in the above link. A much more descriptive set of features, `presentation_style`  objects, are available that can be used to describe recommended presentation of entire classes of annotations or captures segments without requiring individual specification.

I believe this should be in-tree so as to make these field names common across applications and encourage adoption and convention.

The current implementation is blocked by #146 but this can be changed to work without it.

### DIFF
--- a/extensions/presentation.sigmf-ext.md
+++ b/extensions/presentation.sigmf-ext.md
@@ -1,0 +1,190 @@
+# The `presentation` SigMF Extension Namespace v1.0.0
+
+This document defines the `presentation` extension namespace for the Signal
+Metadata Format (SigMF) specification. This extension namespace contains objects
+to help define how SigMF metadata should be presented graphically to users.
+
+All fields in the `presentation` extension are optional, thus any SigMF
+application using this extension is not required to implement any particular
+feature. It is RECOMMENDED that applications adhering to the `presentation`
+extension use default values in this document.
+
+## 0 Datatypes
+
+The `presentation` extension defines the following SigMF datatypes:
+
+|type|long-form name|description|
+|----|--------------|-----------|
+|`color`|Hexadecimal Color String|String representing a 24 bit color (with optional alpha value) in hexadecimal form; either "#RRGGBB" or "#AARRGGBB".|
+
+### 0.1 The `color` datatype
+
+The `color` datatype is a string which MUST be either 6 or 8 hexadecimal
+characters long prepended by a "#". It MAY be specified in capital or lower case
+characters and applications MUST handle either. The alpha channel is optional,
+and while compliant software MUST be able to parse either format presentation of
+the alpha channel by visualization software is RECOMMENDED but OPTIONAL. If
+software is unable to render the alpha channel then it SHOULD ignore the alpha
+value for foreground objects and ignore background objects entirely.
+
+ARGB format was selected over RGBA for compatibility with QColor objects, RGBA
+color specification is invalid and will result in incorrect display of content.
+
+For greatest compatibility, it is RECOMMENDED that this value default to
+`#00ffffff` (fully opaque white) for foreground features and `#ffffffff` (fully
+transparent) for background features.
+
+### 0.2 The `styles` object
+
+The `style_dict` object is a JSON object containg style information by short
+form name. This is a simple mechanism to prevent every annotation from defining
+these fields multiple times.
+
+## 1 Global
+
+The `presentation` extension adds the following fields to the `global` SigMF
+object:
+
+|name|required|type|description|
+|----|--------|----|-----------|
+|`capture_default`|false|`presentation_style`|Default style to be used for captures when a style is not specified.|
+|`capture_styles`|false|object|Color of the fill of the annotation, this is generally used with defined alpha channel.|
+|`annotation_default`|false|`presentation_style`|Default style to be used for annotations when a style is not specified.|
+|`annotation_styles`|false|object|Color of the foreground captures segment display features.|
+
+Both `capture_styles` and `annotation_styles` objects have keys that represent
+capture segments or annotation `label` fields they apply to and values
+consisting of `presentation_style` objects, see examples in Section 4 below.
+
+### 1.1 The `presentation_style` object
+
+The `presentation_style` object is a JSON object where each value is an object with
+the following form:
+
+|name|required|type|default|description|
+|----|--------|----|-------|-----------|
+|`display_type`|true|`display_type`||Color of the fill of the annotation, this is generally used with defined alpha channel.|
+|`line_color`|false|color|`#00ffffff`|Color of the foreground captures segment display features.|
+|`line_width`|false|double|1.0|Width of the line (in pixels) surrounding the annotation, fractional values are permitted.|
+|`fill_color`|false|color|`#ffffffff`|Color of the background, generally used with significant alpha.|
+|`comment`|false|string||User comment field, any string is fine.|
+
+The `display_type` field is required for `presentation_style` objects.
+
+### 1.1.1 The `presentation_style` `display_type` field
+
+The `display_type` field can have the following values:
+
+|value|description|
+|----|-------|
+|`none`|Do not apply special presentation to this segment, all other fields are ignored.|
+|`box`|Apply a box of `line_color` and `line_width` enclosing defined samples, with background `fill_color` (if defined).|
+|`diagonal`|Fill the boxed segment with the specified `fill_color` and apply diagonal `line_color` lines to the foreground.|
+|`grayscale`|This captures segment should be rendered completely desaturated, other fields are ignored.|
+|`point`|Single circle of diameter `line_width` at the center frequency and first sample position. Valid only for annotations.|
+|`impulse`|Line covering the entire spectral range at the first sample position. Valid only for annotations.|
+
+## 2 Captures
+
+The `presentation` extension adds the following fields to the `captures` SigMF
+object:
+
+|name|required|type|default|description|
+|----|--------|----|-------|-----------|
+|`style`|false|[presentation style](presentation.sigmf-ext.md#the-presentation-style-field)|{}|Style to use for the segment.|
+
+## 3 Annotations
+
+The `presentation` extension adds the following fields to the `annotations`
+SigMF object:
+
+|name|required|type|default|description|
+|----|--------|----|-------|-----------|
+|`style`|false|[presentation style](presentation.sigmf-ext.md#the-presentation-style-field)|{}|Style to use for the annotation.|
+|`color`|false|`color`|`0x00ffffff`|Simple specification for the annotation box color.|
+
+The `color` and `style` fields SHOULD NOT both be provided. In the event of an
+overdetermined specification for color, the first specification found below
+should be used to determine the appropriate presentation style:
+
+1. `annotations->presentation:style`
+1. `annotations->presentation:color`
+1. `global->presentation:annotation_styles`[`annotations->core:label`]
+1. default (white box)
+
+## 4 Examples
+
+Example usage of the simple presentation specification, in this case specifying
+different box colors (red, then green) in the annotations directly:
+```json
+  ...
+  "annotations": [
+    {
+      "core:sample_start": 1000,
+      "core:sample_count": 100,
+      "core:label": "signal1",
+      "core:freq_lower_edge": 119000000,
+      "core:freq_upper_edge": 120000000,
+      "presentation:color": "#00ff0000"
+    },
+    {
+      "core:sample_start": 5000,
+      "core:sample_count": 100,
+      "core:label": "signal2",
+      "core:freq_lower_edge": 119000000,
+      "core:freq_upper_edge": 120000000,
+      "presentation:color": "#0000ff00"
+    },
+  ]
+```
+
+How to use `capture_styles` or `annotation_styles` to specify the same two
+annotations above:
+```json
+  "global": {
+    ...
+    "presentation:capture_styles": {
+      "VALID": {
+        "display_type": "none",
+        "comment": "Valid capture segments have no unique presentation."
+      },
+      "INVALID": {
+        "display_type": "diagonal"
+        "line_color": "#80ffff00",
+        "line_width": "3.0",
+        "fill_color": "#c0000000",
+        "comment": "Invalid sections are shaded slightly grey with yellow diagonal lines."
+      }
+    },
+    "presentation:annotation_styles": {
+      "signal1": {
+        "display_type": "box",
+        "line_color": "#00ff0000",
+        "comment": "Box signals labeled 'signal1' in RED."
+      },
+      "signal2": {
+        "display_type": "box",
+        "line_color": "#0000ff00",
+        "comment": "Box signals labeled 'signal1' in GREEN."
+      }
+    }
+  },
+  "captures": [...],
+  "annotations": [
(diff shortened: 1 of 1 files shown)


extensions/presentation.sigmf-ext.md
```diff
@@ -0,0 +1,239 @@
...
+
+The `presentation` extension defines the following SigMF datatypes:
+
+|type|long-form name|description|
+|----|--------------|-----------|
+|`color`|Hexadecimal Color String|String representing a 24 bit color (with optional alpha value) in hexadecimal form; either "#RRGGBB" or "#AARRGGBB".|
+
+### 0.1 The `color` datatype
+
+The `color` datatype is a string which MUST be either 6 or 8 hexadecimal
```
gmabey, line 23: Did you consider also allowing strings from this well-defined set of color names? https://www.w3.org/TR/SVG11/types.html#ColorKeywords

extensions/presentation.sigmf-ext.md
```diff
@@ -0,0 +1,239 @@
...
+and while compliant software MUST be able to parse either format presentation of
+the alpha channel by visualization software is RECOMMENDED but OPTIONAL. If
+software is unable to render the alpha channel then it SHOULD ignore the alpha
+value for foreground objects and ignore background objects entirely.
+
+ARGB format was selected over RGBA for compatibility with QColor objects, RGBA
+color specification is invalid and will result in incorrect display of content.
+
+For greatest compatibility, it is RECOMMENDED that this value default to
+`#00ffffff` (fully opaque white) for foreground features and `#ffffffff` (fully
```
gmabey, line 35: I think you might have your alpha's switched on the two color codes used in this one line.

extensions/presentation.sigmf-ext.md
jacobagilbert: I think this would be ver nice overall but I am not sure if this is burdensome on applications implementing this. Interestingly the merge request with inspectrum does support this (thanks to QT's native support for it).

Open to any thoughts here, especially situations this would be a problem.

extensions/presentation.sigmf-ext.md
jacobagilbert: Yep, thanks. Fixed.

extensions/presentation.sigmf-ext.md
gmabey: Another thought (I'm not strongly advocating this, but I do like it) is to separate the color and the alpha into two separate fields.  That would eliminate the RGBA vs ARGB issue, and simultaneously allow for alpha-izing named colors.

Inasmuch as the `.sigmf-meta` files are supposed to be "human readable", and inasmuch as I still have a hard time visually decoding hex color strings, named colors would be ... more understandable.

jacobagilbert: human readability is a strongly beneficial reason to do this. Not sold on separate alpha but i see your use case...will think more. could be an optional that overrides AARRGGBB....

bhilburn:
I'm in favor of this, and looks good. I'd say get it going, but deprecate your version to something < v1.0.0. Get feedback from @miek and others and target a v1.0.0 release with next SigMF minor release.

bhilburn (approved):
Deprecate the version

gmabey:
That would be duplicate information, no? (Having alpha in 2 places)

Also, a separate alpha field has a very natural default value of 1.

miek:
I've thought about this a bunch and, while I really like the idea of being able to encode some visual information, I worry that this implementation will lead to people creating recordings that only work well with their favourite tool/theme/colour-map. For example, I plan to add options for other colour maps to inspectrum at some point and presentation parameters for annotations created now could be unreadable on other colour maps.

I'd much rather see some way to encode intent rather than specific colours/styles, and then visualisation tools can interpret that in a way that works with their theme. Perhaps the naming scheme from standard logging levels could work, with fatal/error/warn/info/etc.? There could also be a way to define extra named groups, to group similar packets together for example.

777arc:
> I've thought about this a bunch and, while I really like the idea of being able to encode some visual information, I worry that this implementation will lead to people creating recordings that only work well with their favourite tool/theme/colour-map. For example, I plan to add options for other colour maps to inspectrum at some point and presentation parameters for annotations created now could be unreadable on other colour maps.
> 
> I'd much rather see some way to encode intent rather than specific colours/styles, and then visualisation tools can interpret that in a way that works with their theme. Perhaps the naming scheme from standard logging levels could work, with fatal/error/warn/info/etc.? There could also be a way to define extra named groups, to group similar packets together for example.

Great point, although I would rather see integers starting at 0 than fatal/error/warn/info, that way we don't have to all agree what each code means, plus the RFML folks will love integers.  And it solves the group problem.  Would be nice to at least standardize on green and red, e.g. have 0 = green and 1 = red and then past that it's whatever the tooling thinks looks good?

jacobagilbert:
@miek Your input here is great, sorry I sat on this for so long. I'll definitely keep this in mind the next time I'm wondering why one of my inspectrum PRs hasn't been merged yet LOL!

I think your proposal makes a ton of sense and I'd like to figure out a way to capture that effectively. I have run into exactly the concern you raise repeatedly in my playing around with colored annotations in just the single default inspectrum colormap. Im wondering if we can identify a small set of features or annotation groups users would want to allow a tool to color. I think it needs to be a small enough set that its reasonable to find a set of annotation styles that coexist with a given colormap and are distinguishable. 

I like taking inspiration from log levels, heres my first-pass take on this:

```
styles = {
  ALERT,    // should be visible immediately, regardless of the view settings of the dataset
  ERROR,    // should quickly draw attention and be styled in such a way to indicate a problem (e.g.: red)
  WARNING,  // upon inspection it should be obvious that this area may require additional inspection
  INFO,     // this is the standard annotation style
  DEBUG,    // this annotation is of lower importance than standard annotations
  USER_A,   //  \
  USER_B,   //  |
  USER_C,   //   > these styles should be visibly distinct but no relative importance is implied
  USER_D,   //  |
  USER_E    //  /
}
```
A number of questions:
1. Does 10 styles seem like a lot?
2. Should there be **more** user classes?
3. Should the ability to specify suggestions or explicit colors still be included?

---

Also, as an aside:

> For example, I plan to add options for other colour maps to inspectrum

🙌 🙌  I like to hear this, viridis cant come soon enough :) - [Tammojan's suggestion in 204 of using tinycolormap](https://github.com/miek/inspectrum/pull/204) seems like a pretty easy way to get `Parula, Heat, Jet, Turbo, Hot, Gray, Magma, Inferno, Plasma, Viridis, Cividis, Github, Cubehelix, and HSV` for the cost of adding a drop down, removing `Jet` from that list and adding a little interface glue.

777arc:
@jacobagilbert is going to remove some of the fields that havent been used, in order to get a minimal version merged in

devnulling (approved):

devnulling:
@Teque5 @jacobagilbert ping on this fellow

marcusmueller (changes requested):

dkozel (commented):
There are a few typos/errors that I've pointed out here. 

However the much more substantive issue that I'll object to is the use of display styling in place of meaning. I agree there should be a mechanism for indicating that certain elements should be visually styled in a particular (though maybe not absolute) way. 

Introducing visual styling to an element is/should be marking an annotation or capture as being part of some semantic group. The proposed `capture_tags` in #146 introduces the concept of tags as a way of linking meaning to elements and could be extended to `annotation_tags`.

Currently an annotation's `label` does this in part, but there are several reasons not to use it for display styling. There is room for adding more flexibility to users to mark multiple properties of an annotation (without adding a full schema extension) and tags are a viable route there. Label is a mature field which has developed specific meaning to many users and altering it's utility now would be invasive. Third, the decision of what property should be used to decide the display styling should be left as a standalone decision for users.

I was initially sold in on the idea of linking display styles to the `_tag(s)` value, but now am wondering if that becomes overloading functionality. Marcus raised the comparison with HTML element classes and CSS styling. I'm leaning towards having a single `presentation:class` optional per annotation or capture. In the `global` name space it should be required to define
```
"presentation:classes" : [
  {
    "presentation:class": "group1",
    "presentation:description": "Packets sent by users in the Demo group"
  }, ...
]
```

Then either in a separate `.sigmf-style` file or in the global section, have `presentation:style` definitions which refer to the classes. This decouples meaning from the styling details, allowing easy theming.

I can't think of a situation where the styling of a single annotation or capture should be manually set within the scope of the annotation. Styling an element should be used to convey meaning. If meaning is being conveyed it should be defined. If it's defined it should be done in a reusable way. Once it's defined in the global space it can be used for a single element or many. The ability to embed styling information inside of individual objects should be removed.

Thoughts?

I could see reversing the direction of reference such that a style file (or section) defines a set of available styles, and the presentation classes declare what style to use.

Part of my thinking is that the selection/definition of classes is very end-user/application specific, and the styling is also likely to be tool/environment specific.

The Recording would need to declare what Style file it is intended to be used with. Marcus brought up the idea of having the field support a URI, so styles could optionally be remotely hosted, distributed alongside the recordings, or embedded in tools.

I want to like the idea of defining a default set of style names, like the Log Levels and other attributes @miek mentioned, as long as there's a way of defining new styles and distributing them.

marcusmueller:
I like @dkozel's approach [here](https://github.com/sigmf/SigMF/pull/234#pullrequestreview-3075083424), though I have to say that a separate `.sigmf-style` file sounds like something I'd avoid; especially since, well, citing myself, CSS already exists, and reinventing that wheel in JSON certainly isn't a great idea.

I'd lean towards having a global attribute that refers to the URI (not URL) of a separately available style (in whichever format) rather than inclusion in metadata; comparable to, say, how one references an XML namespace. Representation instructions are *not* metadata!

Teque5:
We (@dkozel, @777arc, @KelseyCreekSoftware) just had a long discussion on this topic in the monthly SigMF meeting. We discussed CSS, Matplotlib colormaps, non-web interfaces, hex colors, and more.

## Conclusions
- We want to [KISS](https://en.wikipedia.org/wiki/KISS_principle).
- We want to leave color science and color choices to other folks.
- We want to support web and non-web applications.
- We want to support theming and flexibility where possible.

### Decision
1. The presentation extension should have something that indicates the class or index of the annotation color from [0, N).

### Other ideas
2. The presentation extension should have something that indicates the rank, level, or class of the annotation from [0, N).
3. The presentation extension could support an optional array of named/hex colors that color_index are taken from, or perhaps a defined colormap.

## I Suggest
```markdown
|name|required|type|description|
|----|--------|----|-----------|
|`color_index`|true|int|Color index from colormap. Default 0.|
|`style_index`|false|int|Style index where `0` is highest importance, similar to markdown levels where increasing value is of lower and lower visibility. Default 0.|
```
or we could consolidate both keys into something called `index` that just describes the color and style and rank all in one attribute.

@777arc is going to work up a new PR on this.

_edit: removed color suggestions.

777arc:
Yeah I think consolidating into 1 index is best, to keep it very simple, we can make a note that the first index, 0, could correspond to the signal of interest or friendly signals, and that the second index, 1, could correspond to interferers or adversarial signals, that way if software wants to use green and red for those categories it knows exactly what to do. 

Side note, I think the different log level meanings don't translate intuitively into signals in a spectrogram, I like "ALERT" to bring attention to a signal but none of the other ones feel good.

777arc:
Actually, before I open a new PR, does it make sense to create an entirely new extension if all I'm proposing is adding an optional field to `annotations` called `color_index` which is an integer starting at 0 and goes as high as you want?  It would be up to the software to decide how to map to color, but we would have the suggested mapping using the following color scheme, which is just [matplotlib's default aka tab10](https://matplotlib.org/stable/users/explain/colors/colormaps.html#qualitative) but with green and red as 0 and 1 respectively:

0 #2ca02c (green) - can be used for signal of interest, friendly signals, signals of high importance, etc
1 #d62728 (red) - can be used for interferers, adversarial signals, etc
2 #1f77b4 (blue)
3 #ff7f0e (orange)
4 #9467bd (purple)
5 #8c564b (brown)
6 #e377c2 (pink)
7 #7f7f7f (grey)
8 #bcbd22 (mustard)
9 #17becf (cyan)
<suggested colors only go up to 9 but the color_index can go as high as you want>

There could be example python/javascript/etc code somewhere that provides the mapping so it's easy for folks

So the point here is to not have a standard SigMF way to define styling, but instead leave it up to the software, and have a very light method that folks can choose to use to categorize annotations by color, either N classes, or SOI/interferer style.  Someone might even just want a way to mark extra important signals so they only ever use 0 and no other levels.

marcusmueller:
Because I've been pretty vocal in chat, so that I can't come back later and complain :D :

I agree with https://github.com/sigmf/SigMF/pull/234#issuecomment-3657051755 and https://github.com/sigmf/SigMF/pull/234#issuecomment-3657112824 , i.e. annotations are classes/indices (not sure why classes need to be numeric, not all classes are orderable and you can't always know which classes you want before, so strings do sound nicer in some contexts, but I'm not going to complain. It feels much more useful when looking things up to have an integer.).

Nice work, thank you for hashing this out.

Teque5:
> Actually, before I open a new PR, does it make sense to create an entirely new extension if all I'm proposing is adding an optional field to `annotations` called `color_index` which is an integer starting at 0 and goes as high as you want? It would be up to the software to decide how to map to color, but we would have the suggested mapping using the following color scheme, which is just [matplotlib's default aka tab10](https://matplotlib.org/stable/users/explain/colors/colormaps.html#qualitative) but with green and red as 0 and 1 respectively.

I think any choice you make to specify a specific color is not a **good idea**. Barring any discussion of cultural color symbolism and accessibility concerns, I think colors are left to the GUI and the designers.

If folks are interested in color theory, [this talk related to the introduction of the matplotlib `viridis` colormap is mandatory watching on this topic.](https://www.youtube.com/watch?v=xAoljeRJ3lU)

I went back and edited by suggestion to remove tab10.

I also don't mind if we make this an optional key in annotations, it's more likely to be used that way.

777arc:
Yeah we can leave out the suggested color mapping, I agree software will use its own scheme based on the spectrogram colormap.

GassiusODude:
I recently wrote a JavaFX spectral analyzer that supports user selected color schemes for annotations: https://github.com/GassiusODude/spectral_analyzer/

My take is that you can't satisfy everyone's subjective choice of color scheme and the choice of colors can change due to the presentation goal and the heatmap used (as others have noted).  I don't think the specification needs a "presentation" field.  Instead, it should be based on the application and the desired presentation.

I went with applying a color scheme tied to the label of the annotation to start.  For instance, I might want to highlight different modulations if I'm working on something that specifically targets one type.  Or if doing RF fingerprinting, I might want to label each burst based on the expected emitter.  As a user, I can represent this by changing my labels.  The current SigMF already supports custom fields in the annotations.  The display application can then just support accessing these custom fields (like modulation, emitter, RF standard).  SigMF just needs to worry about better representing the data.  The application then picks up the slack to chose how to present the material based on user's need.

PR #239, opened 2022-06-02
deprecate tuples as collection Recording pointers

jacobagilbert:
This is done in favor of JSON objects, avoiding nested vectors (prohibited by flatbuffers) and also permitting expansion within this object.

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -139,7 +139,7 @@ distribution in a SigMF `Archive`.
                  ║                     ║ │
                  ║   SigMF Recording   ╟─┘
                  ║                     ║
-                 ║ (base SigMF object) ║
+                 ║ (base SigMF Object) ║
                  ╚══════════╤══════════╝
                             │
                             │ comprises
@@ -242,7 +242,7 @@ whitespace, line-endings, EOF characters).
 
 Complex samples MUST be interleaved, with the in-phase component first (i.e.,
 `I[0]` `Q[0]` `I[1]` `Q[1]` ... `I[n]` `Q[n]`). When `core:num_channels` in the
-Global object (described below) indicates that the Recording contains more than one channel,
+Global Object (described below) indicates that the Recording contains more than one channel,
 samples from those channels MUST be interleaved in the same manner, with
 the same index from each channel's sample serially in the Recording. For
 example, a Recording with two channels of `ri16_le` representing real-valued
@@ -259,8 +259,8 @@ SigMF metadata fundamentally takes the form of key/value pairs:
 "namespace:name": value,
 ```
 
-Metadata field names in the top level `global` object, `captures` segment
-objects, or `annotations` objects MUST be of this form. All fields other than
+Metadata field names in the top level `global` Object, `captures` segment
+Objects, or `annotations` Objects MUST be of this form. All fields other than
 those at the top level which contain a `:` delimiter SHALL only use letters,
 numbers, and the `_` character; all other characters are forbidden. Field names
 MUST NOT start with a number and MUST NOT not be C++20 or Python 3.10 keywords.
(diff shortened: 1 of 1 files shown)


sigmf-spec.md
```diff
@@ -663,22 +663,42 @@ this field be capable of displaying up to 20 characters.
...
-in an ordered list via a JSON array.
+called a `collection`. The Collection object associates specific Recordings
+together through key/value pairs which link metadata keys and
+`SigMF Recording Objects` that point to Recordings containing actual data. The
+following rules apply to SigMF Collections:
 
 1. The Collection Object MUST be the only top-level object in the file.
-1. The Collection Object MUST only contain SigMF key/value pairs.
-1. Keys in the Collection Object MUST use SigMF Recording Tuples to reference 
+1. The Collection Object MUST only contain SigMF key/value pairs provided by the
```
bhilburn, line 672: I would remove this rule from this list, and rather add it to the rules about Compliance later in the spec.

sigmf-spec.md
```diff
@@ -663,22 +663,42 @@ this field be capable of displaying up to 20 characters.
...
+
+Or as a tuple (deprecated) of the same form. The [`name`, `hash`] order of the
+tuple is mandatory.:
+
+```JSON
+  ["example-channel-0-basename", "2e747e0694082a7983b92406b5b394971806988ed0862bd216188e5583f59eda]
+```
+
+Tuples will be removed in SigMF version 2.0, so JSON Objects are RECOMMENDED.
+Additional optional user fields MAY be added to `SigMF Recording objects` if
```
bhilburn, line 699: `objects` should be capitalized

sigmf-spec.md
```diff
@@ -663,22 +663,42 @@ this field be capable of displaying up to 20 characters.
...
-1. SigMF Recording Tuples MUST take the form of `["N", "hash"]`, where `N` is the 
-   base-name of a SigMF Recording and `hash` is the SHA512 hash of the Recording's 
-   Metadata file `N.sigmf-meta`.
-1. If a value contains multiple SigMF Recording Tuples, they MUST appear in a JSON array.
+1. SigMF Recording Objects MUST contain both a `name` which is the base-name of
+   a SigMF Recording and a `hash` which is the SHA512 hash of the Recording
+   Metadata file `[base-name].sigmf-meta`.
+1. If a value contains multiple SigMF Recordings, they MUST appear in a JSON array.
+
+`SigMF Recording Objects` reference the base-name of the SigMF Recording and the
```
bhilburn, line 681: This should be a sub-section with a table describing the required fields in the Object, like we do elsewhere in the spec. Reading it, I realize we should have previously done this for the tuple version, too, but neglected it at the time.

bhilburn (changes requested):
@jacobagilbert - I like this proposed solution. Good idea / design. My comments are all copy related, not design related.

bhilburn:
Bumping, looks like no changes since last review

sigmf-spec.md
jacobagilbert: Don't disagree....but this suggestion feels unrelated to this PR.

jacobagilbert:
@bhilburn thanks for the ping. I think I have addressed most of the feedback. I also did a quick scan through this for `object` --> `Object` language and upgraded a number of uses to the proper `Object` form.

777arc (approved):

bhilburn (approved):
Minor language comment, otherwise this PR is great and is the best way forward for this issue IMO.

sigmf-spec.md
```diff
@@ -689,35 +690,68 @@ The following names are specified in the `core` namespace for use in the `collec
...
+            }
         ]
     }
 }
 ```
 
+#### SigMF Recording Objects
+
+`SigMF Recording Objects` reference the base-name of the SigMF Recording and the
+SHA512 hash of the Metadata file, and can be specified as a JSON Object:
```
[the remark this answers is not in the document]
jacobagilbert: Yep, this is correct will fix and merge.

Thanks!

jacobagilbert:
Issues addressed, multiple approvals. Merging.

PR #240, opened 2022-06-03
specify true north in spatial extension

jacobagilbert:
also improves examples


gmabey:
I approve, all good changes.

PR #249, opened 2022-11-21
Added Matrix chat info and monthly call to README

777arc:


jacobagilbert (approved):
So long as you are OK w/ your email in our readme, LGTM :)

PR #250, opened 2022-11-21
CI Pipeline for Python Package

777arc:
Runs unit tests on each push, and will email you if there's an error.  Not sure if it blocks a merge, that might be in repo settings.

image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,19 @@
+name: Continuous Integration
+on: [push]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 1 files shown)


jacobagilbert (approved):

jacobagilbert:
@777arc Can we have 3.6, 3.8 and 3.10 pipelines? Thinking specifically about the recent issue with #253

777arc:
I thought we said the python package was supposed to work under python 3.6, which would catch #253 assuming there's a unit test that covers that part. Python is pretty good about backwards compatibility across major releases so 3.6 should be the only thing we need to test under.

PR #253, opened 2022-11-30
sigmf_hash: use file_digest

willcode:
This could speed up hashing of large files.

Signed-off-by: Jeff Long <willcode4@gmail.com>

### DIFF
--- a/sigmf/sigmf_hash.py
+++ b/sigmf/sigmf_hash.py
@@ -27,8 +27,6 @@ def calculate_sha512(filename):
     """
     Returns sha512 of filename
     """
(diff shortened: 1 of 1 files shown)


jacobagilbert (approved):
Good change 👍

daniestevez:
Hi, I've just noticed this change because it fails in Python 3.8.10. Apparently [`hashlib.file_digest`](https://docs.python.org/3/library/hashlib.html#hashlib.file_digest) is new in Python 3.11. Probably we would like to fall back to the old way of computing hashes in older Python versions.

willcode:
Ugh, sorry. I thought I checked that.

Issue #208, opened 2021-12-28
questioning the utility vs. hassle of the N/ directory thing


gmabey:
Hello,

I'm reading the spec, in the section "Rules for SigMF Archive files:" where it reads:
```
3. The Archive MUST contain the following files: for each contained Recording with some name
    given here meta-syntactically as N, files named N (a directory), N/N.sigmf-meta, and N/N.sigmf-data.
4. The Archive MAY contain a .sigmf-collection file in the top-level directory.
```
and I'm starting to question the `N/` directory thing.  It seems to me that the original intent may have been to "keep the family together" when un-taring an archive at the command line, but the Collection thing isn't consistent with this goal.

Now, don't get me wrong, I *like* collections, and I *like* keeping families together.  However, I'm just now looking at using `archivemount` to give me access to datasets without copying them, and the extra directory seems silly in that context (the context of having all of the contents of an Archive accessed through a new base directory).

So, here's my "contribution question":

How is the `N/` directory better than recommending a simple bash function/script that looks something like this:
```bash
unarchive_sigmf() {
    FILE_BASENAME=`basename $1`
    ARCHIVE_NAME=${FILE_BASENAME%.sigmf}
    mkdir -p $ARCHIVE_NAME
    tar --directory=$ARCHIVE_NAME -xf $1
    echo "SigMF Archive has been extracted to ./$ARCHIVE_NAME/"
    echo "and have a _great_day_."
}
```
Probably should add some error reporting to the above, but you understand the general approach, right??

I'm also starting to feel that the `N/` directory thing being a "MUST" conflicts with "Collections rule 3": "The sigmf-collection file MUST be either in the same directory as the Recordings that it references, or in the top-level directory of an Archive (described in later section)."  The contradiction comes in that, if the `.sigmf-data` and `.sigmf-meta` files MUST be in an `N/` directory, how could a `.sigmf-collection` file possibly be simultaneously "in the same directory as the Recordings that it references" since each recording would have to be in its own `N/` directory?  Unless of course there's only one recording in a collection, which I guess I don't see as a requirement anywhere, but I kinda thought that a plurality of recordings was intrinsic to the concept of a collection.

Preemptive (and hypothetical) rebuttal responses:
[...]

PR #256, opened 2022-12-03
updating language on sigmf archives

jacobagilbert:
This specifies that subdirectory use of recording files is not required and suggests that users not use this structure for archives.

It also specifies that at least one SigMF Recording must be present in an archive, which i think is a requirement.

@gmabey your input/review is appreciated!

Closes https://github.com/gnuradio/SigMF/issues/208

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -166,16 +166,16 @@ distribution in a SigMF `Archive`.
 
 Rules for all files:
 1. All filetypes MUST be stored in separate files on-disk.
-1. It is RECOMMENDED that filenames use hyphens to separate 
-   words rather than whitespace or underscores.
+1. It is RECOMMENDED that filenames use hyphens to separate words rather than
+   whitespace or underscores.
 
 Rules for SigMF Metadata files:
 1. A Metadata file MUST only describe one Dataset file.
 1. A Metadata file MUST be stored in UTF-8 encoding.
 1. A Metadata file MUST have a `.sigmf-meta` filename extension.
-1. A Metadata file MUST be in the same directory as the Dataset file
-   it describes.
-1. It is RECOMMENDED that the base filenames (not including file extension) of 
+1. A Metadata file MUST be in the same directory as the Dataset file it
+   describes.
+1. It is RECOMMENDED that the base filenames (not including file extension) of
    a Recording's Metadata and Dataset files be identical.
 
 Rules for SigMF Dataset files:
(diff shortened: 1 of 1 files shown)


sigmf-spec.md
```diff
@@ -200,10 +200,15 @@ Rules for SigMF Collection files:
...
 1. The Archive MUST use the `tar` archive format, as specified by POSIX.1-2001.
 1. The Archive file's filename extension MUST be `.sigmf`.
-1. The Archive MUST contain the following files: for each contained Recording
-   with some name given here meta-syntactically as `N`, files named `N` (a
-   directory), `N/N.sigmf-meta`, and `N/N.sigmf-data`.
-1. The Archive MAY contain a `.sigmf-collection` file in the top-level directory.
+1. The Archive MUST contain at least one SigMF Recording.
+1. The Archive MAY organize SigMF Recordings into subdirectories such that for
+   each contained Recording with some base name `N`, individual files are stored
+   within the directory `N` (e.g.: `N/N.sigmf-meta`, and `N/N.sigmf-data`). The
```
gmabey, line 206: I suggest that all references to `N/N` be removed, and just say that any directory structure can be used -- the only requirement is that the `.sigmf-meta` file be in the same directory (and have the same name stem) as the `.sigmf-data` file.
For Non-Conforming Datasets, I don't see a good reason for there to be any correlation between the directory/name of the `.sigmf-meta` file and the `dataset` field.

Alternative language:
"The Archive MAY organize SigMF Recordings into subdirectories, but the `.sigmf-meta` and the `.sigmf-data` files must be in the same directory (except for Non-Conforming Datasets, see the description of the `dataset` field below)."

sigmf-spec.md
```diff
@@ -200,10 +200,15 @@ Rules for SigMF Collection files:
...
+   use of subdirectories within an archive is NOT RECOMMENDED and may be
```
gmabey, line 207: Why NOT RECOMMENDED? Why not just let users be free to use whatever *internal* directory structure they want to? Seems to me that (from the application side) for `subdir1/subdir2/a_recording.sigmf-{meta,data}` treating the entire `subdir1/subdir2/a_recording` section as the recording "name" will be adequate for most purposes: opening files (after extraction), identifying contents within unextracted archives, and adding files to new archives.

sigmf-spec.md
```diff
@@ -200,10 +200,15 @@ Rules for SigMF Collection files:
...
+   deprecated in future versions of SigMF.
+1. The Archive MAY contain a `.sigmf-collection` file in the top-level directory,
```
gmabey, line 209: I like the old version of this rule.

gmabey (commented):
I think rules and language that don't contribute to the use cases should be treated suspiciously.

sigmf-spec.md
jacobagilbert: This was updated to ensure visibility on the fact that sigmf-collection files MUST NOT exist with the subdirectory permission above.

sigmf-spec.md
jacobagilbert: Permitting arbitrary directory structures places a somewhat obnoxious burden on parsing applications to do directory traversal through the archive, and it obfuscates how many Recordings are present in an archive. The benefit of permitting this seems negligible at best. Permitting more than one Recording in an Archive is a bit OBE now that SigMF has Collections anyway.

sigmf-spec.md
jacobagilbert: See above. Users are still free to manage Recordings, Collections, and SigMF Archives however they please and can use tar archives or compressed archives to store data outside of what is explicitly codified by this specification. The main purpose of canonical SigMF Archives is to specify a simple, single, directly usable file (that should generally contain either one Recording or one Collection). To that end they should be as straightforward as possible.

sigmf-spec.md
gmabey: If the user is accessing the data without extracting the contents of the archive, then I see no difference in usage -- that is to say, I don't see anything obnoxious about that. If the user *wants* to double the disk usage by extracting an archive (yeah, beats me why you would want to do that <wink>) then the `--strip-components` switch to `tar` can eliminate the directory structure entirely and just leave you with files at the base of the tree.

That being said, I'm not worried about it if I don't prevail in this -- there are more important issues than this. I was trying to gently push back on the (IMHO unnecessary) extra stipulation that recording files *should* go in the base of the tar tree -- in the interest of a shorter spec.

jacobagilbert: We definitely want to support accessing data without extracting archive contents, your examples there are excellent.

I'll leave this open for a bit and get more input. I wont unilaterally force a solution either way. I just am sensitive to anything that makes sigmf application development more difficult.

jacobagilbert:
Discussion above is marked as resolved, @gmabey can you take a look at the updates? I believe these are generally as we discussed yesterday with the possible exception of the last rule I added.

sigmf-spec.md
jacobagilbert: Entirely removed in favor of a more permissive rule.

gmabey:
hmm -- I tried to add a comment to the Pull Request but am disappointed it didn't show up in this discussion.  I like it all, but I do suggest that "user-defined directory structures" be replaced with "arbitrary directory structures".

jacobagilbert:
Strange. I actually went back and forth and will switch to 'arbitrary'. Thanks!

jacobagilbert:
Closes #208

bhilburn (approved):
Collections replaces the old approach.

jacobagilbert:
Approvals from Ben and Glen, merging.

PR #261, opened 2022-12-23
Add formal SigMF Collection support

### LINKED ISSUE #252: Add `ci16_le` and `collection` examples to README
the current fc32 example is much more straightforward due to numpy having a native type for this. should add a ci16 example since its not obvious.

Also should include an example on how to create a collection.

jacobagilbert:
This creates a base SigMFMetafile class that Recording and Collection objects inherit from for common functions. This also separates some of the Collection object keys that were somewhat haphazardly thrown into the SigMFFile class despite having no real use there yet.

It does not handle archiving or schema based validation of collections yet, those common functions should probably be added to the base class. Example code to test these new features are in the [README on this branch](https://github.com/gnuradio/SigMF/tree/add_collection_handling#create-and-save-sigmf-recordings-and-collections-from-a-numpy-arrays)

Also updates README to provide an example of collections - Closes #252

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,9 @@ for adx, annotation in enumerate(annotations):
     samples = signal.read_samples(annotation_start_idx, annotation_length)
 ```
 
-#### Write a SigMF file from a numpy array
+#### Create and save a Collection of SigMF Recordings from numpy arrays
+
+First, create a single SigMF Recording and save it to disk
 
 ```python
 import datetime as dt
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -160,7 +160,9 @@ for adx, annotation in enumerate(annotations):
     samples = signal.read_samples(annotation_start_idx, annotation_length)
 ```
 
-#### Write a SigMF file from a numpy array
+#### Create and save a Collections of SigMF Recordings from numpy arrays
```
777arc, line 163: should it be Collection?

README.md
```diff
@@ -201,7 +203,51 @@ meta.add_annotation(100, 200, metadata = {
...
+    }
+)
+collection.tofile('example_zeros.sigmf-collection')
+```
+
+The SigMF Collection and its associated Recordings can now be loaded like this:
+
+```python
+from sigmf import sigmffile
+collection = sigmffile.fromfile('example_zeros')
```
777arc, line 248: Is there a way to get a list of stream_names in the collection, if so this would be the perfect place to show how, since a lot of folks might know they are opening a collection but not know the names of each stream

sigmf/sigmffile.py
```diff
@@ -798,36 +938,48 @@ def fromarchive(archive_path, dir=None):
 
 def fromfile(filename, skip_checksum=False):
     '''
-    Creates and returns a returns a SigMFFile instance with metadata loaded from the specified file.
-    The filename may be that of either a sigmf-meta file, a sigmf-data file, or a sigmf archive.
+    Creates and returns a returns a SigMFFile or SigMFCollection instance with metadata loaded
```
777arc, line 941: extra "returns a"

README.md
jacobagilbert: Done, with a simple list comprehension to get all the sigmf files also.

README.md
jacobagilbert: Good catch

sigmf/sigmffile.py
jacobagilbert: 🤦

777arc (approved):

PR #262, opened 2022-12-23
updating license for python module to LGPLv3

### LINKED ISSUE #254: License Considerations for Python Module
Currently this repository (housing the specification AND the example python module) is licensed under Attribution-ShareAlike (CC BY-SA 4.0) however even the originator of this license [discourages its use for code](https://creativecommons.org/faq/#can-i-apply-a-creative-commons-license-to-software).

I believe the specification being CC licensed makes sense, however we should probably consider a GPL variant for the python module. Given this is simply a reference implementation to access an open source specification, I don't think there are any real concerns with this code being used commercially... however we may want some copyleft guarantees.

jacobagilbert:
This specifies the LGPLv3 License for the python module instead of the unsuitable CC license the spec uses. It is not ideal to have the spec and python code in the same repository under different licenses, but that will be addressed post v1.1.0 release when they are split into their own repos.

Closes #254 

~~`DONT MERGE` for now.~~

### DIFF
--- a/sigmf/__init__.py
+++ b/sigmf/__init__.py
@@ -1,22 +1,8 @@
-# Copyright 2021 GNU Radio Foundation
+# Copyright: Multiple Authors
 #
-# Permission is hereby granted, free of charge, to any person obtaining a copy
-# of this software and associated documentation files (the "Software"), to deal
-# in the Software without restriction, including without limitation the rights
-# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-# copies of the Software, and to permit persons to whom the Software is
-# furnished to do so, subject to the following conditions:
+# This file is part of SigMF. https://github.com/gnuradio/SigMF
 #
-# The above copyright notice and this permission notice shall be included in
-# all copies or substantial portions of the Software.
-#
-# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-# SOFTWARE.
+# SPDX-License-Identifier: LGPL-3.0-or-later
 
 __version__ = '1.1.0'
 
--- a/sigmf/archive.py
+++ b/sigmf/archive.py
@@ -1,22 +1,8 @@
-# Copyright 2021 GNU Radio Foundation
+# Copyright: Multiple Authors
 #
-# Permission is hereby granted, free of charge, to any person obtaining a copy
-# of this software and associated documentation files (the "Software"), to deal
-# in the Software without restriction, including without limitation the rights
-# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-# copies of the Software, and to permit persons to whom the Software is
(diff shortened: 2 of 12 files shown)


Teque5 (approved):

bhilburn:
I'm good with the LGPL license, but the header swapped in to the code files in this PR is actually the GPLv3 header, not the LGPL header?

jacobagilbert:
Thanks @bhilburn you are correct - I got this [notice from here](https://www.gnu.org/licenses/gpl-howto.html) but failed to read the entire section 🤦

jacobagilbert:
I think this is mergable, but no real rush. Removing the `DONT MERGE` label.

bhilburn (changes requested):
@jacobagilbert - I see that you changed it to an SPDX identifier, but it looks like you used the GPL string instead of the LGPL string?

```text
SPDX-License-Identifier: GPL-3.0-or-later
```

jacobagilbert:
Good grief. Sorry, thanks Ben - FIXED!

dkozel (changes requested):
I agree with LGPL being a reasonable license for a reference implementation of our open spec. The compelling immediate desire is to increase the usage of the open SigMF file format over proprietary or less useful formats, not to compel users of the SigMF file format to release their higher level applications.

Just a note, there is no GNU Radio **Foundation** to hold the copyright. Nor was their in 2021... "GNU Radio project" could be used (though technically I don't think the project can hold copyright since it isn't a formal legal entity). But as intent is a big part of court interpretations of copyright cases I'd still go with "GNU Radio project". @bhilburn?

jacobagilbert:
Yeah, this is probably a good time to get that correct as well. This project has never required copyright assignment either, so its not clear what we can/should do regarding that notice.

@bhilburn i think you are the expert opinion here :)

bhilburn:
Great catch, @dkozel.

Making it "GNU Radio Project" communicates intent, but also indicates assignment, which as @jacobagilbert notes, we have never / will never do.

Just make it "Copyright Multiple Authors", which is fine

jacobagilbert:
OK, thats better. Thanks!

sigmf/archive.py
```diff
@@ -1,22 +1,8 @@
-# Copyright 2021 GNU Radio Foundation
+# Copyright 2021, 2022 Multiple Authors
```
Teque5, line 1: For the year do we need to keep appending? Can we just make it 2023 and keep it clean?

jacobagilbert: Not really sure actually... a quick google search seems to indicate that the first year of publication is all that is necessary.

bhilburn: Including a year actually hasn't been necessary since 1989, but everyone does it anyway.

In short, I don't care, and it doesn't really matter, so do whatever you like / is easiest 🤣

jacobagilbert: `Copyright: Multiple Authors` it is

jacobagilbert:
I think this is ready. If I can get a few final reviews ill merge ahead of a release.

gmabey:
Looks good to me,

sigmf/__init__.py
```diff
@@ -1,22 +1,8 @@
...
-# all copies or substantial portions of the Software.
-#
-# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-# SOFTWARE.
+# SPDX-License-Identifier: LLGPL-3.0-or-later
```
bhilburn, line 5: Typo in this identifier

bhilburn (changes requested):

sigmf/__init__.py
jacobagilbert: 😳

PR #263, opened 2023-01-05
Archive reader simplification

### LINKED ISSUE #244: SigMF archive error reading data file
I think that there is an error opening sigmf archive files. The compliant dataset file is no recognized as a valid file because the relative path of the tar member is used https://github.com/gnuradio/SigMF/blob/1eef44a67636ab15078e7eecc00144b9e18cdaa3/sigmf/sigmffile.py#L779

jacobagilbert:
Closes #244 

@gmabey please take a look at this when you are able.

### DIFF
--- a/sigmf/archive.py
+++ b/sigmf/archive.py
@@ -25,7 +25,7 @@ import shutil
 import tarfile
 import tempfile
 
-from . import error
+from .error import SigMFFileError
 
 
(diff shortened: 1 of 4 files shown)


jacobagilbert:
Not happy with this... closing and will reopen with a different fix.

PR #264, opened 2023-01-06
use SigMFArchiveReader for fromfile()

### LINKED ISSUE #244: SigMF archive error reading data file

jacobagilbert:
Closes #244

Also cleans up SigMFFileError imports/usage.

### DIFF
--- a/sigmf/archive.py
+++ b/sigmf/archive.py
@@ -11,7 +11,7 @@ import shutil
 import tarfile
 import tempfile
 
-from . import error
+from .error import SigMFFileError
 
 
 SIGMF_ARCHIVE_EXT = ".sigmf"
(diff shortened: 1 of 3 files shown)


jacobagilbert:
@gmabey does this seem like a reasonable way to address 244? I think we have accumulated a bit of architectural debt in the python module and it we can discuss that in the future (there is a circular import issue here which is why the import is inline) but that can be done once we get a functional release up.

gmabey:
Yes, looks very reasonable to me, although I have not yet actually tried it.  And -- I appreciate those cleanups too.

gmabey:
I'm getting a `Metadata in archive did not .validate()!` that I didn't expect, but there's no detail.  It looks like @jacobagilbert had that part commented out in #263.

Actually I've had that error pop up for a while, I just kept hoping it would go away when those validate PRs were finally merged.  Is anyone else seeing that message?

jacobagilbert:
Ah yea i should probably add that. `validate()` no longer returns a bool, it throws an exception if it fails. Thats a bug.

jacobagilbert:
Concurrence on the call today, merging.

PR #269, opened 2023-01-17
separating the python module from the spec

jacobagilbert:
The Python module has been broken out here https://github.com/gnuradio/sigmf-python

I think we are ready to do this.


777arc (approved):

Teque5 (approved):
Are the extensions going to live here too?

jacobagilbert:
Yep, that's the intention.

jacobagilbert:
Two approvals, full speed ahead.

PR #270, opened 2023-02-03
spatial: add range-rate to bearing object

daniestevez:
This adds `range_rate` and `range_rate_error` fields (in m/s units) to the bearing object. These indicate the time derivative of the range field and its uncertainty or error.

We're planning on using range-rate information for GNSS applications, and adding it to the spatial extension bearing object seems most natural, since it already has a range field. A range-rate field seems applicable in other contexts, specially when range-rate is measured via Doppler. Some examples that come to mind are radar and spacecraft navigation.


jacobagilbert (approved):

jacobagilbert:
No objections, is sandboxed to an extension, and is a good idea. Merging.

PR #273, opened 2023-03-10
Add DVB-S2 extension

igorauad:
Hi SigMF maintainers,

I have been organizing my DVB-S2 IQ recordings using SigMF and have decided to document a DVB-S2 extension for it, which is included in this PR. I would be happy to refine it if you find it interesting for the extensions kept in the repository.

Some comments:
- I realized it would be helpful to have `enum` and `union` types for some of the parameters, so I added these. Please let me know if I missed an alternative canonical approach that would fulfill the same purpose.
- I see other extensions have numbered markdown sections (like "1 Global", etc.), but I think numbering in markdown sections is often error-prone, so I didn't number the sections. Please let me know if you object to that.

Regards

### DIFF
--- a/extensions/dvbs2.sigmf-ext.md
+++ b/extensions/dvbs2.sigmf-ext.md
@@ -1,0 +1,178 @@
+# DVB-S2 Extension v1.0.0
+
+## Abstract
+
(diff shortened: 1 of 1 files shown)


jacobagilbert:
Hello @igorauad  - thank you for submitting this! I'd like to redirect this PR to the new https://github.com/sigmf/community-extensions repository, where I hope to move almost all signal-specific, and many other extensions.

Teque5:
Closing PR, assuming author will resubmit to the community-extensions repository.

PR #274, opened 2023-03-23
[WIP] Reworded num_channels portion and mentioned Collections

777arc:
Reworded num_channels to make it clear it was intended for real signals and Collections are intended for multiple channels of complex signals, it has proven to be an area of confusion and most importantly we don't want folks to assume num_channels is *the* way to capture multiple RF channels off a receiver.

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -202,6 +202,24 @@ Rules for SigMF Archive files:
    arbitrary directory structures, but the SigMF files within the Archive MUST
    adhere to all rules above when the archive is extracted.
 
+When representing real-valued recordings that contain more than one channel, such as from a multichannel data acquisition
+device, use of the interleaved multichannel format via the 
+field in the Global Object (described below) may be more appropriate.
+For example, a Recording with two channels of `ri16_le` representing real-valued
+audio data from a stereo Recording and here labeled `L` for left and `R` for
+right, the data MUST appear as `L[0]` `R[0]` `L[1]` `R[1]` ... `L[n]` `R[n]`.
+
+When representing multiple channels of complex valued data that are related, such as IQ data from multichannel RF receiver,
+you can either use the SigMF [Collection](#sigmf-collection-format) format  or the `core:num_channels` attribute of the global section.
+Using the Collection format, which is  ultimately multiple single channel recordings,
+you'll be able to open a single channel of your Collection in common SDR tools.
+On the other hand, using `core:num_channels` will allow streaming to/from SigMF Archives,
+and there is only one file to deal with, but because the multiple
+channels of IQ will be interleaved, it is no longer a standard binary IQ file, and may have less
+support using common SDR/DSP tools.
+If your start_time's are not aligned, Collections is the only option.
+Certain data structures and operations in memory may benefit from the channels already being interleaved, when using num_channels.
+
 ### SigMF Dataset Format
 
 There are four orthogonal characteristics of sample data: complex or real, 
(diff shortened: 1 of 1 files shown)


gmabey:
Hrm -- I used a RF receiver once that produced multi-channel packets of complex data. Seems like a very efficient way to implement beam forming algorithms: you've got all of the channels right there, and time synchronized.

Why not allow multi-channel complex recordings? Is the only motivation to prevent users from getting confused? Or, is utility and efficiency a higher, nobler way? For a user fluent in numpy N-D-isms, seems like N=2, dtype=complex32 isn't really a stretch.

Why do you say that `num_channels` isn't the "proper" way to capture multiple RF channels off a receiver?  What is improper about it?  How then is storing multiple channels of real samples proper?

I certainly have encountered the case where channels were not necessarily time synchronized, and I *thought* that a collection would very nicely cover that case.

I myself really appreciate the potential to store multiple channels of complex data with a single time stamp.

777arc:
Didn't use the word proper, but the point here is to make sure folks dont read the spec and assume that num_channels is the way to represent multiple rx channels.  In the end its a tradeoff between the number of files you have to deal with, and ease of ingesting the files into other software.  It's not hard to deinterleave, but for a lot of folks its going to be the difference between being able to directly use certain software to analyze their recordings and not.

gmabey:
I apologize -- I thought for sure I saw the word "proper" in your comment somewhere.

So, the motivation for this proposed change is ... compatibility?  That is, to make it more likely for software that ignores the `num_channels` quantity to use data in a recording?

777arc:
Yeah pretty much, even though right now there's not a ton of tooling around SigMF, over time there will be more, and some fraction isn't going to support num_channels (since it was intended for real-valued signals like audio and DAQs anyway)

777arc:
Also if I step back and think about examples of RF recordings that have multiple channels, what comes to mind is stuff like multi-element arrays, and satellite ground stations where you get H and V polarizations.  So Sentinel 1 for example releases their raw RF signals to the world and they have H and V as separate files.  For multi-element arrays you're going to be doing some form of MIMO DSP and its likely a custom piece of software.

jacobagilbert:
I'd like a few wording tweaks here @777arc - I don't think it is right to recommend either format for multichannel real valued systems; the best format is going to depend on the application.


> **_When_** representing multiple channels of **_complex valued data_** that **_are related, such as multiple channels of IQ data_**, the Collections format is recommended. **_When representing_** real-valued recordings that contain more than one channel, **_such as from a multichannel data acquisition device, using the interleaved multichannel format and the_** `core:num_channels` field in the Global Object (described below) **_may be more appropriate._**

Let me know what you think.

777arc:
Great suggestion, I've updated the paragraph.  Do we want to still include that L/R example at the end or should we go deeper into num_channels related stuff further down the doc?

gmabey:
I do like this new version "better" but I'm still surprised that there's interest in differentiating between real values and complex values.  What's so different about them that warrants saying anything at all? "A shorter spec is a better spec." -- *attributed to Cicero, circa 75 BC* LOL

jacobagilbert:
The point of this recommendation (as I see it) is to encourage compatibility with tools like Inspectrum and IQEngine. SigMF was created with the intent of managing raw IQ recordings from radios, and that remains the focus; "a binary file of IQ or RF samples" is the only example in our readme for a reason. Suggestions also get to the point of "make it easy to use" which is a stated objective -- less pain means a better user experience.

> A shorter spec is a better spec.

I definitely agree with this, but there is a rule that supersedes that one:

> A good spec provides exactly one way to accomplish something

And because that is not always possible, I think recommendations like this are suitable.

jacobagilbert:
I don't care too much for the recent edits and I would like to revisit the original point of this:

Its not obvious how one should store multichannel complex valued data in SigMF, and almost every tool that exists to manipulate such data (inspectrum, baudline, gnuradio, UHD, soapy, iq_engine, sceptre, spike, gqrx, hdsdr, sdrangel, countless govt tools, etc etc) accept IQ data per-channel. SigMF should, since we don't hate our users and want to make their lives easier, direct new people toward **Collections** for this use case.

Nothing about this would _prevent_ or stop anyone from implementing this however they see fit, and there are some niche reasons to use single-file recordings. This is merely intended to help people who otherwise don't know to do this in the standard way.

Issue #275, opened 2023-04-05
No Json Schema Definition


Nepomuceno:
There is no Json Schema https://json-schema.org/ that we could use to validate a metadata file.

The solution would be to create a json schema file that would encompass preferable all the core values but also the extensions if possible.

jacobagilbert:
Validation is indeed an area we need help with.

Are there other problems with the schema besides not having extension fields in it?

Nepomuceno:
Yes there are problem on data validation for the string and numerical fields for example the datatype does allow any string to be input and the same are true for the uint fields that do not limit the min anx max for all the scenartios.

I should have a PR adding those fix by the end of the day today... My idea was to put this PR in a spec folder in this repo but I am open to suggestions.

777arc:
I think the lack of a maintained json schema has been a hole, and it should live within this repo, we'll just have to determine the cleanest way for the python package to get a copy of it (e.g. submodule this repo?).

Question- for the schema associated with each extension, is there a clean way to have them in separate files?  I don't think it would be sustainable to have one large schema that includes the extensions.

jacobagilbert:
To add to Marc's point on individual extension schemas, adding them to a monolithic schema is unappealing since it makes it harder for people to validate just what they care about, and more importantly to add their own schema files for their own extensions.

Nepomuceno:
You can use $ref in order to extend JSON schemas that way if you were using extensions you could include the $ref and use all of those schemas and for someone developing their own schema they can use $ref with their own subschemas to include the properties they want 

https://json-schema.org/understanding-json-schema/structuring.html

PR #276, opened 2023-04-09
Add spec file for metadata.

Nepomuceno:
First step trying to improve the specs at #275

## What it does 
Add the basic Spec file fore metadata.
This is inspired in the file of https://github.com/sigmf/sigmf-python/blob/main/sigmf/schema-meta.json

## Assumptions

these assumprions need to be verified

- [ ] There is a minimum of one channel that it is specified in the files.
- [ ] Sample retes are always positive
- [ ] The only version current cupported by this schema it is 1.0.0
- [ ] Frequencies can be negative

## What it is missing 

In his pr it is still missing the implementation of the extensions but space it is given to them.

I did not have many sample files to tesst the validation of the schema if someone could help me with that would be great.

### DIFF
--- a/spec/1.0.0/schema-meta.json
+++ b/spec/1.0.0/schema-meta.json
@@ -1,0 +1,325 @@
+{
+    "$id": "https://github.com/sigmf/SigMF/spec/1.0.0/schema-meta",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "default": [
+        "global",
+        "captures",
+        "annotations"
+    ],
+    "required": [
+        "global",
+        "captures",
+        "annotations"
+    ],
+    "title": "Schema for SigMF -meta file.",
+    "type": "object",
+    "properties": {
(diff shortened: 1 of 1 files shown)


777arc:
1. There is a minimum of one channel that it is specified in the files, correct, an empty captures array would negate the need for the recording to exist imo
2. Sample rates are always positive, correct
3. Yep it's OK if the latest release (1.0.0) is the only one supported, although we should talk about what the process is when a new release is added, e.g. we should archive the older schemas in a nice way 
4. Frequencies can be negative <currently in discussion in sigmf matrix chat, will update>

777arc:
Is it possible to make an example of having a separate schema file for an extension, e.g. it could be a 1 field extension but just showing how the $ref stuff will look, and then people who have extensions can use it as an example for making their schema file?  Not sure how much work that would be.

777arc:
closing because this will be included in the #292

PR #281, opened 2023-04-19
Allowing geolocation in captures and making it preferred

### LINKED ISSUE #278: Geolocation in captures/global/annotations
We had a request that geolocation also be allowed in captures for moving receivers that get a periodic location update.  This brought up the argument that it might as well only be in captures, just like frequency, the only downside would be backwards combability issues.  Tooling designed to only have 1 coordinate for a Recording (e.g. a maps interface on a tool) can just pull it from captures[0].

On a related note, it would be nice to have a way for an annotation to contain a geolocation of the emission associated with that annotation (the transmitters location, not receiver).  This could be results from TDOA or it could be "label data" known beforehand, etc.   This one seems a lot simpler but should it be part of an extension or the core spec?

777arc:
Deprecate it in global in v2 and remove in v3?

Closes #278

### DIFF
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -470,6 +470,7 @@ Segment Objects:
 | `header_bytes`  | false    | uint   | The number of bytes preceding a chunk of samples that are not sample data, used for NCDs.  |
 | `frequency`     | false    | double | The center frequency of the signal in Hz.                                                   |
 | `datetime`      | false    | string | An ISO-8601 string indicating the timestamp of the sample index specified by `sample_start`.|
+| `geolocation`   | false    | GeoJSON `point` Object | The location of the Recording system. |
 
 **The `sample_start` Field**
 
--- a/sigmf-spec.md
+++ b/sigmf-spec.md
@@ -472,7 +472,7 @@ Segment Objects:
 | `header_bytes`  | false    | uint   | The number of bytes preceding a chunk of samples that are not sample data, used for NCDs.  |
 | `frequency`     | false    | double | The center frequency of the signal in Hz.                                                   |
 | `datetime`      | false    | string | An ISO-8601 string indicating the timestamp of the sample index specified by `sample_start`.|
-| `geolocation`   | false    | GeoJSON `point` Object | The location of the Recording system.|
+| `geolocation`   | false    | GeoJSON `point` Object | The location of the Recording system at the start of this Captures segment.|
 
 **The `sample_start` Field**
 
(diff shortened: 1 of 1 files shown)


aromanielloNTIA:
This seems like a more limiting change to the spec than the one proposed in #237, although maybe I am misunderstanding the use case here.

777arc:
They are very similar, it's one of those tradeoffs where if it's part of an extension you'll get less support in tooling and less folks using it.  I guess my thought was, moving it from global to captures doesn't add complexity to the core spec, but gives that little bit of flexibility some might need.

jacobagilbert:
This is certainly more limited than what I proposed in #237 but I do think this adds value and I don't think we have a great way of addressing 237 comprehensively currently. This does allow a sort of strange way where each GPS point is logged in a new capture in a metadata_only sigmf file that is part of a collection but I don't think thats a good solution.

aromanielloNTIA:
Maybe there's a split of 3 different ways for handling geolocation metadata:

Case 1 (recordings by static receiver at a single location): Use the global geolocation field and no others.
Case 2 (recordings by static receiver at multiple locations): Use the method added by this PR, where each capture has its own geolocation information.
Case 3 (recordings by mobile receiver): Still needs a proper solution, the primary topic of #237

This seems a little fragmented, but I think it's OK. I would not like to see the global geolocation field deprecated, since static receivers performing multiple captures in the same location is a reasonable use case, and this change would require duplication of the geolocation metadata for each capture. An example of this would be a static receiver recording IQ samples at multiple frequencies. Perhaps the global field could remain, and the captures field could take precedence if it exists?

jacobagilbert:
I think I agree that migrating the `global` scope `geolocation` object to `captures` scope is the right way to address a variety of use cases. it still does not cover the situation where you want to capture true time-varying position within a SigMF Collection so we can leave 237 open. Open to ideas over there though!

jacobagilbert (approved):

sigmf-spec.md
```diff
@@ -498,6 +472,7 @@ Segment Objects:
 | `header_bytes`  | false    | uint   | The number of bytes preceding a chunk of samples that are not sample data, used for NCDs.  |
 | `frequency`     | false    | double | The center frequency of the signal in Hz.                                                   |
 | `datetime`      | false    | string | An ISO-8601 string indicating the timestamp of the sample index specified by `sample_start`.|
+| `geolocation`   | false    | GeoJSON `point` Object | The location of the Recording system.|
```
jacobagilbert, line 475: I think we should change this to:

> The location of the Recording system **at the start of the Capture segment**.

Teque5 (approved):
Changes make sense to me, however I think if this is merged we should finalize a v1.1.0 release.

PR #282, opened 2023-04-19
Added geolocation to spatial extension within annotations

777arc:
To capture the estimated or known location of a transmitter/emission being annotated

### DIFF
--- a/extensions/spatial.sigmf-ext.md
+++ b/extensions/spatial.sigmf-ext.md
@@ -213,7 +213,7 @@ for simplicity because many applications only require azimuth, but the
 measurement error is needed. Only one of these SHOULD be used for an individual
 annotation; if both are provided the `signal_bearing` object has priority.
 
-`geolocation` uses the same GeoJSON `point` Object as in the `core` spec, but here it corresponds to an emission or signal, not the receiving platform's position.  
+`geolocation` uses the same GeoJSON `point` Object as in the `core` spec, but here it corresponds to an emission or signal, not the receiving platform's position.
 For example, this field could be used to store results of TDOA performed on an emission, or it could be used to label a known position of a transmitter.
 See `core` spec for more details on GeoJSON `point` Objects.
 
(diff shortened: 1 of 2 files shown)


extensions/spatial.sigmf-ext.md
```diff
@@ -203,15 +203,20 @@ object:
 |----|--------|----|-----|-----------|
 |`signal_azimuth`|false|double|degrees|Azimuth in degrees east of true north associated with the specific annotation.|
 |`signal_bearing`|false|[bearing](spatial.sigmf-ext.md#01-the-bearing-object)|N/A|Bearing associated with the specific annotation.|
+| `geolocation`  | false    | GeoJSON `point` Object | N/A | The location (either known or estimated) of the transmitter corresponding to this annotation |
```
jacobagilbert, line 206: lets call this `signal_geolocation` or `emitter_geolocation` to avoid confusion with the [Annotations Scope `latitude` and `longitude` fields](https://github.com/sigmf/SigMF/blob/sigmf-v1.x/sigmf-spec.md#annotations-array).

jacobagilbert (approved):
This seems OK but the next thing that will be desired is a way to capture uncertainty... is that something you have a good idea for?

extensions/spatial.sigmf-ext.md
jacobagilbert: Yes, it is indeed confusing. Originally these fields were designed to hold the _system_ location and were annotations-scoped to allow changing position. This predates me a bit, but I think everyone agrees it was not the ideal choice.

These have been deprecated since 0.9 and will be formally removed in 2.0

777arc:
@gmabey or @Nepomuceno can you give this a sanity check before we merge it?

777arc:
also @jacobagilbert  do we bump v1.0.0 to like 1.1.0?

jacobagilbert:
yea, we should.

Nepomuceno:
Looks good. That would definitely be used by us.

777arc:
ok @jacobagilbert i believe it's ready to merge

argilo:
The extension version number in the example should probably be updated too.

PR #292, opened 2023-08-22
[WIP] Specs now live as a JSON Schema, and the markdown is rendered

777arc:
A few months ago we started to create a JSON Schema out of our specs, in order to produce a more formal set of specifications for the JSON meta file.  Then I started playing around with tooling that auto generates specs based on a JSON Schema.  It worked, but created a really verbose markdown file, so I switched over to a short custom script that essentially just builds the tables and lists of fields.  I also switched to using YAML, which surprisingly is a perfectly valid way to represent a JSON Schema, because it allows us to include multi-line descriptions and code blocks and such while JSON requires you to use \n's.  So it's still a JSON Schema, we can generate a sigmf.schema.json file if we really want to, but for now the idea is to have this JSON Schema YAML file be the one place that the specs live, and the markdown will be generated from it.  I'll play around with creating a GitHub action that generates the markdown file whenever the YAML is edited. 

We definitely need eyes on this one, as far as going through the generated markdown to see if everything is valid, because it's too much to visually diff.  And if anyone is an expert with JSON Schemas or knows someone, we can really use help refining the schema. 

For reference, commit https://github.com/sigmf/SigMF/commit/6353dcce65f8b269395b95818716aecf742b6d71 was used to produce this first version so once it gets approved, I'll update it with whatever PRs have merged since this point

### DIFF
--- a/md-generator.py
+++ b/md-generator.py
@@ -1,0 +1,43 @@
+import yaml
+from py_markdown_table.markdown_table import markdown_table
+
+sigmf_version = 'v1.0.0'
+data=yaml.safe_load(open('sigmf.schema.yaml', 'r'))
+
+def gen_table(d):
+    table = []
+    for key, value in d["properties"].items():
+        newdict = {"Field": '[`' + key.replace('core:','') + '`](#' + key.replace('core:','') + ')'}
+        newdict["Required"] = "Required" if key in d["required"] else ""
+        newdict["Type"] = value.get("type", "MISSING")
+        newdict["Default"] = str(value.get("default", ""))
+        description = value.get("description", "")
+        indx = description.find('.')
+        newdict["Short Description"] = description[:indx].replace('\n','') # short description, which is up to the first period
+        table.append(newdict)
+    return markdown_table(table).set_params(quote=False, row_sep = 'markdown', padding_weight='right').get_markdown()
+
+def gen_fields(d):
+    fields_str = ''
+    for key, value in d["properties"].items():
+        fields_str += "### " + key.replace('core:','') + '\n\n'
+        fields_str += value["description"] + '\n\n'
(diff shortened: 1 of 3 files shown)


jacobagilbert:
This is really, really cool 👍 

I think for this to be integrated a couple things need to happen:

- CI action to ensure the schema matches the rendered MD file
- One schema to rule them all (per namespace...) - this means replacing the schemas for the [core metadata](https://github.com/sigmf/sigmf-python/blob/main/sigmf/schema-collection.json) and [collections](https://github.com/sigmf/sigmf-python/blob/main/sigmf/schema-collection.json) in the `sigmf-python` module.
- decide how to handle extension namespaces

jacobagilbert:
Blocked by #296

Teque5:
> Blocked by #296

Just approved #296.

jacobagilbert:
@Teque5 what do you think about this concept in general? Goal here is to get the technical details of the spec in exactly one place that can be used for software validation AND generating the spec tables.

jacobagilbert:
I am hopeful we can get the markdown file cleaned up a bit, especially with the benefit of #296... what exists today certainly needs to be fixed a bit, e.g.:

```
datatype
The SigMF Dataset format of the stored samples in the Dataset file.

examples: ['ri16_le']
default: cf32_le
pattern: ^(c|r)(f32|f64|i32|i16|u32|u16|i8|u8)(_le|_be)?$
type: string
```

But we should be able to get close to what we have now in the MD. The goal should be that we don't increase any burden for _understanding_ the spec. Making changes to SigFM will definitely be a little more complicated...

777arc:
Yep this change shouldnt increase the barrier to entry at all, because the produced markdown should nearly match the existing, if anything it should lower it for dev-oriented folks who like working with schemas to understand a spec

Issue #297, opened 2023-09-14
Rename default branch, resolve versioning


Teque5:
Now that the spec has been separated from the [python module](https://github.com/sigmf/sigmf-python), the versions are now separate too. There are a few outstanding issues in this repo that I'd like to resolve that were remnants from the before-times:

* rename default branch from `sigmf-v1.x` to `main`
* rename mis-named tags (`SigMF-v1.0.0`) to `v1.0.0`
* create new release for `v1.2.0`. Most recent release is `v1.0.0`, but most recent tag is `v1.1.0`.
* update the spec example version to match current release

If another maintainer signs off on these changes I can do this myself.

Do we want to move the logo to a separate repo?

777arc:
I really like the idea of using "main" as the default branch name.  I think logo is OK to stay in this repo, although an SVG of the logo would be pretty sweet if anyone can manage that =).  I would suggest waiting for the json schema to get merged in before we cut a release, we can bump the version of the json schema along with the title in the markdown at the same time.

Teque5:
> I think logo is OK to stay in this repo

It's a `.sigmf` file as an example where we can keep the metadata up to date w/the spec. 

> although an SVG of the logo would be pretty sweet if anyone can manage that =)

There are (7) `svg` versions of the logo [in the zip file on this page](https://github.com/sigmf/SigMF/tree/sigmf-v1.x/logo). I think what you are suggesting is that we add `logo-scope-color.svg` to the repo, which may make sense since it's the most useful version. When I first added the `png`, I did that since GitHub wouldn't render an `svg` in a markdown file, but they added that feature in 2022. I'll make a PR.

Teque5:
As there are no objections I'm going to proceed:

- [ ] renaming the base branch to `main` and tags the appropriate commits for other versions.
- [ ] create a PR that updates the spec to `1.2.0`.
- [ ] Create a v1.2.0 Release (with notes)
- [ ] Update `sigmf-python` to reflect spec version increment

777arc:
Sounds good, I'll update my PR at some point

Teque5:
All tasks complete.

PR #298, opened 2023-10-02
Clarify the format of core:version

argilo:
Some of the examples in the specification use `1.0.0`, while others use `v1.0.0`. The definition of `core:version` does not specify the exact format.

GNU Radio's SigMF Sink uses `1.0.0`, and it's the only version accepted by IQEngine's validator, so I would assume that was intended to be the correct value.

I've updated the examples and added a clarification to the definition of `core:version`.

### DIFF
--- a/extensions/spatial.sigmf-ext.md
+++ b/extensions/spatial.sigmf-ext.md
@@ -434,7 +434,7 @@ element recording:
 ```JSON
 {
     "collection": {
-        "core:version": "v1.0.0",
+        "core:version": "1.0.0",
         "core:extensions" : [
             {
             "name": "spatial",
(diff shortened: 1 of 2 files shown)


gmabey:
The python implementation also omits the "v" in its `__init__.py` (currently "1.1.3").

gmabey:
Not to imply that the python-sigmf version is synonymous with the version of the standard ... just a data point.

Teque5:
The most recent tag in this SigMF repo is `SigMF-v1.1.0` so shouldn't it be changed in the spec to `1.1.0`?

This issue is also related to #297, where I'd like to fix the tags and push a `1.2.0` release.

argilo:
> The most recent tag in this SigMF repo is SigMF-v1.1.0 so shouldn't it be changed in the spec to 1.1.0?

Maybe, but that seems like a separate issue that could have its own pull request. I used 1.0.0 because that's what is written everywhere in the document currently.

argilo:
I updated this to explicitly state that the format is `X.Y.Z`.

/cc @jacobagilbert

jacobagilbert:
Thanks @argilo - I'd like to remove the "currently this is..." part though. Mostly because I don't want to need to update this every release. Reasonable? I'll get this merged.

argilo:
Sounds fine. Removed.

Teque5 (approved):

PR #299, opened 2023-10-05
Access Control Extension

romanziske:
Hi,

I am currently working on a feature to control access to SigMF files. For this purpose I have defined the Access Control Extension. With this extension you can define an owner of the SigMF dataset. The owner can decide if the dataset is public or which people or groups can view and edit the metadata. This extension is inspired by the Google Drive sharing functionality and can be useful when used by a management system like IQEngine.

### DIFF
--- a/extensions/access_control.sigmf-ext.md
+++ b/extensions/access_control.sigmf-ext.md
@@ -1,0 +1,84 @@
+# Access Control Extension for SigMF v1.0.0
+
+The Access Control Extension is designed to manage access to SigMF files, providing mechanisms for access permissions and control.
+
+## 1 Global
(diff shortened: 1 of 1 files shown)


jacobagilbert:
Hi @romanziske - I want to thank you for taking the time to submit this. You are proposing a solution to a very real problem here, however I think this is blurring SigMF into the domain of access control which is almost surely better handled using a more typical method such as filesystem permissions or another approach focused on access mangement.

I do find the notion of an owner to be useful, but probably not so much that it warrants a distinct field from the `core:author`.

As always you are most certainly welcome to keep and use this extension, and when we have a location for community extensions I'd welcome it there. For now though I don't think this is something we will pull into the canonical extensions.

PR #300, opened 2023-10-06
Clarify optional extensions

argilo:
It wasn't entirely clear to me what it means for an extension to be optional. I assume it means that the application must support the extension in order to parse the recording. This might be the case if the extension defines a new dataset format.

I've updated the definition of `optional` according to my understanding, and also updated the example below. I changed the `spatial` and `antenna` examples to `"optional": true` since I don't believe that these extensions define anything that applications should be required to parse.

Finally, I noticed that the `antenna:hagl` and `antenna:azimuth_angle` fields in the `antenna` example were broken, so I fixed them.

### DIFF
--- a/extensions/spatial.sigmf-ext.md
+++ b/extensions/spatial.sigmf-ext.md
@@ -439,7 +439,7 @@ element recording:
             {
             "name": "spatial",
             "version": "1.0.0",
-            "optional": false
+            "optional": true
             }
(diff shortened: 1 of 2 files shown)


Teque5 (approved):
lgtm

jacobagilbert:
I think this is a reasonable clarification.

PR #306, opened 2024-01-13
swap PNG for SVG logo now that GitHub renders

Teque5:
Now that GitHub will render an SVG in markdown files, replace the low resolution PNG with a nice vector image. Unfortunate `svgz` is not supported, since that one is only 11KB.

Best we can do for now per request in #297 by @777arc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,4 @@
-<p align="center">
-<img src="https://github.com/gnuradio/SigMF/blob/sigmf-v1.x/logo/sigmf_logo.png" width="30%" />
-</p>
+<p align="center"><img src="https://github.com/gnuradio/SigMF/blob/sigmf-v1.x/logo/sigmf_logo.svg" alt="Rendered SigMF Logo"/></p>
 
 # Signal Metadata Format (SigMF)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,8 +1,6 @@
-<p align="center">
-<img src="https://github.com/gnuradio/SigMF/blob/sigmf-v1.x/logo/sigmf_logo.png" width="30%" />
-</p>
+<p align="center"><img src="./logo/sigmf_logo.svg" alt="Rendered SigMF Logo"/></p>
 
-# Signal Metadata Format (SigMF)
```
777arc, line 5: I think removing this will hurt SEO, recommend leaving

777arc:
Is it OK if we leave the png?  Both in case anyone is already referencing it, and just because it's convenient for folks who need a png

Teque5:
Now keeping PNG and SVG; leaving SEO unoptimization.

Teque5:
I recall that the reason we left that long URL in the README was so that when PiPy uses the readme as the doc it still renders correctly.

777arc (approved):

PR #309, opened 2024-02-09
Update capture_details.sigmf-ext.md

### LINKED ISSUE #307: capture_details extension: "attentuation" typo
https://github.com/sigmf/SigMF/blob/a6eb9d7bb05455eeb5cfff75bffe0b64dc827064/extensions/capture_details.sigmf-ext.md?plain=1#L18

The key should be "attenuation" not "attentuation".  It is spelled correctly in the description.  Unfortunately this would not be a backward-compatible change.
If it is decided to not change this to maintain compatibility then the typo should at least be called out in the description, as this can easily cause bugs for generators or parsers that may have correctly spelled "attenuation."

jacobagilbert:
Closes #307


jacobagilbert:
This might cause issues but this was a mistake and is dangerous to leave as-is. Its also likely implemented correctly by end users.

777arc (approved):

PR #317, opened 2024-06-20
Github action to check for docs generator build errors during PRs

777arc:


docs-generator.py
Teque5, line 1: Run `black` and `isort` on this when you get a chance.

.github/workflows/generate_docs_pr.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Apt update
+        run: sudo apt update
+      - name: Install prereqs
+        run: sudo apt install python3-pip inkscape texlive-latex-extra -y
+      - name: Pip installs
+        run: sudo pip install pylatex
+      - name: Install a more recent version of Pandoc than available from apt or pip (to get svg support)
```
Teque5, line 21:
```suggestion
      # Install a more recent version of Pandoc than available from apt or pip (to get svg support)
      - name: Upgrade Pandoc
```

Teque5 (approved):

PR #322, opened 2024-10-21
Clarify NCD filename without directory

### LINKED ISSUE #321: Dataset file in same metadata directory instead of local directory
I think there is an issue with the dataset description in SigMF Specification Version v1.2.2
https://sigmf.org/#ssubsec:dataset

> The Dataset file must be in the local directory, and this string MUST NOT include any aspects of filepath other than the filename.

If I am running a script in `/home/user/`  and I open the `signal.sigmf-meta` file in the directory `/media/usb1/`, the specification is telling us the dataset file has to be in the local directory, ie:  `/home/user/`

I would suggest to modify the dataset specification to specify that the Dataset file must be in the same directory than the .sigmf-meta file.

777arc:
Closes #321


Teque5 (approved):
OK per SigMF monthly call

PR #323, opened 2024-11-18
Add geolocation to captures array

777arc:
redo of #281


Teque5 (approved):
Looks good per discussion.

PR #324, opened 2024-11-29
Add traceability, capture detail, signal, spatial extension as json schema

romanziske:
Changes:
- Added missing schema JSON files to repository
- Updated docs-generator.py to include new schemas

There is still an issue with referenced fields: 
bug_on_refs

We should make it possible to link this to the actual type definition.


777arc:
Awesome ty!

PR #328, opened 2024-12-18
Fix description of 'core:version' and creation of short-descriptions

bbc131:
**Problem**
The short description is generated by truncating the long description at the first period (".").
This doesn't work (for example) for the description of `version` in the `global` object, which contains the format specifier "X.Y.Z", the corresponding output is broken (ends with "X").

**Fix**
With the present commit, the first period followed by a space is used to truncate the short description.
Additionally, a missing period at the end of the description of `version` is added.

### DIFF
--- a/docs-generator.py
+++ b/docs-generator.py
@@ -34,7 +34,7 @@ def gen_table(table, d):
         dtype = value.get("type", "MISSING")
         # default = str(value.get("default", ""))
         longdescription = value.get("description", "")
-        shortdescription = longdescription[: longdescription.find(".")].replace("\n", "")  # short description, which is up to the first period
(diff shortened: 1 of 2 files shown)


Teque5:
Thanks! Seems like a no-nonsense PR, but I don't know where the short description is used so I'll let @777arc approve & merge.

bbc131:
Alright, the short description is used for the tables.

777arc:
Nice catch, and ty for the fix!

777arc:
@bbc131 btw in case you didnt know, you can preview the html and pdf if you go to the github action that runs as part of the PR check, then go to summary and there should be a download at the bottom

777arc (approved):

PR #335, opened 2025-01-31
closes #334

### LINKED ISSUE #334: schema maximum value for samples_start, sample_count, others is awkward
I'm working to use https://github.com/json-schema-org/json-schema-spec to validate SigMF metadata.
I'm finding that the schema-dictated `maximum` value for several properties to be very problematic, and I wonder if there's a good rationale for the current value:
```
18446744073709552000 = 2^64 + 384
```
I would like to suggest that the value `2**63-1` be used instead, as this makes it much easier for libraries like `json-schema-spec` to implement such a constraint.
This **would** cause a breakage for all those files out there with values > 9223372036854775807, it's true. But, given the number of elementary particles in the universe, I don't think that's going to be a problem in practice.
Like, it's one thing to enforce such a limit in python; it's something else to do so in generic C++. The implementation of `json-schema-spec` uses `long` to hold such a constraint, on a JSON element of type `unsigned int` (that's a nlohmann/json thing).

What is the origin of 18446744073709552000 anyway?

gmabey:
reduces maximum constraint to 9223372036854775807 for several properties

### DIFF
--- a/sigmf-schema.json
+++ b/sigmf-schema.json
@@ -75,7 +75,8 @@
           "description": "The index number of the first sample in the Dataset. If not provided, this value defaults to zero. Typically used when a Recording is split over multiple files. All sample indices in SigMF are absolute, and so all other indices referenced in metadata for this recording SHOULD be greater than or equal to this value.",
           "default": 0,
           "minimum": 0,
-          "maximum": 18446744073709552000,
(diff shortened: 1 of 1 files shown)


Teque5 (approved):
Makes sense.

gmabey:
Care to weigh in @777arc?

PR #338, opened 2025-02-04
WIP removed the apparently-unnecessary anyOf blocks

### LINKED ISSUE #337: simplification of sigmf-schema.json
I reworked the schema to not use `"anyOf"` sections, and it seems to work ok still.
However, I suspect that there's something I don't understand about JSON schema; this is my first exposure to it.
As I am using json-schema-validator, this change results in the error descriptions being intelligible ...

gmabey:
Note that (for clarity on the substantive edits) this does not include appropriate indent-level changes.
Closes #337

### DIFF
--- a/sigmf-schema.json
+++ b/sigmf-schema.json
@@ -4,16 +4,26 @@
   "title": "Schema for SigMF Meta Files",
   "description": "SigMF specifies a way to describe sets of recorded digital signal samples with metadata written in JSON. SigMF can be used to describe general information about a collection of samples, the characteristics of the system that generated the samples, features of signals themselves, and the relationship between different recordings.",
   "type": "object",
-  "required": ["global", "captures", "annotations"],
+  "required": [
+    "global",
+    "captures",
+    "annotations"
+  ],
   "properties": {
     "global": {
       "description": "The `global` object consists of key/value pairs that provide information applicable to the entire Dataset. It contains the information that is minimally necessary to open and parse the Dataset file, as well as general information about the Recording itself.",
-      "required": ["core:datatype", "core:version"],
+      "required": [
+        "core:datatype",
+        "core:version"
+      ],
       "type": "object",
       "properties": {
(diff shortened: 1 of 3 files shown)


Teque5:
I think you should include the indent changes here too. I usually do a round-trip in python but I think people also use `jq` for this.

sigmf-schema.json
```diff
@@ -76,7 +91,7 @@
           "default": 0,
           "minimum": 0,
           "!comment": "The maximum value for this property is equal to 2^63 - 1, making it easy to fit into a signed 64-bit integer.",
-          "maximum": 9223372036854775807,
+          "maximum": 9223372036854776000,
```
Teque5, line 94: I don't think you meant to change these. This number is `2**63 + 192`. Was previously max signed integer.

Teque5 (changes requested):

gmabey:
:-O blames `jq`
At least the mystery is solved regarding how the previous value for `maximum` came to be -- it was surely `int64_t(double(2**64-1))`

gmabey:
I suppose *if we were smart* we would use an integer small enough such that `int64_t(double(maximum_value)) < (2**63-1)` ...

Teque5 (approved):

gmabey:
I now notice that the jq-ification rewrote 1e12 -> 1000000000000 which I dislike

gmabey:
Looking for feedback from @777arc (or anyone who can give these changes a spin).
Also, I'm thinking *harder* about that joke/threat I made to suggest a number less than 2**63-1 in an effort to sidestep this issue in the future (jq-ification).  Experimentally I found it to be 9223372036854774784 (tentatively).

777arc:
FYI you can see the rendered docs by going to the "Build docs" check then clicking Summary on the left, then click "compiled_docs", and open sigmf-spec.html.  The core:sample_rate renders the same, everything seems to render the same.

PR #340, opened 2025-05-16
Remove maximum number of channels limitation

daniestevez:
In df53293baec41ab8a2163fa70a84f357ba700d55, the SigMF spec was converted from Markdown to a JSON schema. Minimum and maximum values were added for many fields. In particular, the core:num_channels field got a maximum value of 1000. The Markdown specification did not specify a maximum value for core:num_channels, so in a sense this is a breaking change between v1.2.1 and v1.2.0 of the spec.

The limitation breaks some usage of SigMF to represent channelizer data, such as the output of the a polyphase-filterbank, or (power-detected) spectrogram data.

This removes the limitation by setting the max value to 2**63 - 1, which is the max value used for other fields.

### DIFF
--- a/sigmf-schema.json
+++ b/sigmf-schema.json
@@ -83,7 +83,7 @@
           "description": "Number of interleaved channels in the Dataset file, if omitted this is implied to be 1, for multiple channels of IQ data, it is RECOMMENDED to use SigMF Collections instead of num_channels for widest application support.",
           "default": 1,
           "minimum": 1,
-          "maximum": 1000,
+          "maximum": 9223372036854775807,
           "type": "integer"
         },
         "core:offset": {


daniestevez:
I noticed this when some internal tooling that handles spectrogram data broke with a new release of `sigmf-python`. So I'm interested in how to get the fix through to `sigmf-python`. Since the JSON schema is [copied in sigmf-python](https://github.com/sigmf/sigmf-python/blob/main/sigmf/schema-meta.json), it seems that the appropriate way to proceed is to wait for this PR to be merged, tag a release of SigMF, and then make a sigmf-python PR to copy the new JSON schema. Is that right?

777arc:
Change makes sense!  Yep that's the process, but we can make sure to do it quick in this case, in fact I'll merge and tag the patch release now

777arc (approved):

PR #347, opened 2025-09-18
Add an emitter_label string

daltoncass:
Add a human readable string label for an emitter such as MAC address or an alphanumeric serial number.  The seid field is an int.

### DIFF
--- a/extensions/signal.sigmf-ext.md
+++ b/extensions/signal.sigmf-ext.md
@@ -171,6 +171,7 @@ covered by any of the other fields. Examples include pi/4-DQPSK and GMSK.
 |name|required|type|description|
 |----|--------|----|-----------|
 |`seid`|false|uint|Unique ID of the emitter|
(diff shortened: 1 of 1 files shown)


daltoncass:
Just to clarify, you are suggesting that we use "signal"->"emitter"->"label" instead of  "signal"->"emitter"->"emitter_label", and not suggesting that we just use "core:label" for the emitter serial number/MAC/callsign/etc, correct?

Teque5:
> Just to clarify, you are suggesting that we use "signal"->"emitter"->"label" instead of "signal"->"emitter"->"emitter_label", and not suggesting that we just use "core:label" for the emitter serial number/MAC/callsign/etc, correct?

Yes, just as you have changed it.

Teque5 (approved):

daltoncass:
Does that block merging this?

777arc:
Nope, merging =)

PR #349, opened 2025-12-03
fixed & updated CITATION

### LINKED ISSUE #348: Fix CITATION
If you click on "Cite this repository" in the sidebar it doesn't work correctly. I believe the `CITATION.cff` file is malformed. Should be an easy fix.

Teque5:
Should close #348.

I believe the authors field should probably be updated to the contributors to the repo or `multiple authors` or similar. We don't really have a SigMF ORG yet, but perhaps we should?

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,10 +1,9 @@
 cff-version: 1.0.3
-message: To cite the SigMF specification, please include the following:
 authors:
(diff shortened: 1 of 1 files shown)


Teque5:
@777arc please review

777arc (approved):

PR #350, opened 2025-12-21
Allow global.properties.core:sample_rate to be any number greater than 0

nsbruce:
See https://github.com/sigmf/SigMF/issues/339 for discussion. There are cases where the sample rate of a system is less than 1 second. This change makes use of the [JSON schema `exclusiveMinimum` keyword on ranges](https://json-schema.org/understanding-json-schema/reference/numeric#range).

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,7 +2,7 @@ cff-version: 1.0.3
 authors:
   - name: The GNU Radio Foundation, Inc.
 title: The Signal Metadata Format (SigMF)
-version: 1.2.5
+version: 1.2.6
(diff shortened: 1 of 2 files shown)


dkozel:
This is a nice, minimal, and accurate change. I tested against a minimal sigmf-meta file with a sample rate of 1/(60*60).
https://github.com/dkozel/sigmf-example-files/commit/1b49e6f4622f061d10b21a2156fcc46cb07d622d

Teque5 (approved):

Teque5:
Thanks for the PR!

